PR #2, opened 2025-06-25
Place form subdirectory at the top level

knoepfel:
The `add_subdirectory(form)` call is in `phlex/phlex/CMakelists.txt`, and it should be  `phlex/CMakeLists.txt`.


knoepfel:
Hat tip to @gemmeren for noticing the error.

PR #12, opened 2025-07-29
Remove dependence on boost backtrace

wddgit:
Remove dependence on boost backtrace.  This involves removing the function named ```function_name``` and the function overloads that used it. The calls to ```with``` and ```output_with``` that used those were replace with overloads where the name must be supplied explicitly, usually with a string literal.

I don't understand CMake well so I would suggest reviewing the changes to the CMakeFiles carefully. I was just guessing... Is there documentation somewhere covering CMake usage in phlex that explains how to write a CMakeFile? Is there a way to verify the dependence was removed?

It builds and existing unit tests pass. It seems like this shouldn't need new tests as I am deleting not adding.


knoepfel:
@wddgit, this looks great!  The CMake changes are fine...with the exception of removing the installation of one interface library (I've restored in with commit https://github.com/Framework-R-D/phlex/pull/12/commits/4a943e9bcfb6f833323980030a5a0f618a05bb67).  I confirm that I am able to build this and successfully execute the unit tests.  I've also taken the liberty of executing `clang-format` on the changed files and pushing the changes to the same branch.

By removing the following lines:

```cmake
find_library(BACKTRACE_LIB backtrace REQUIRED)
find_path(BACKTRACE_INC backtrace.h REQUIRED)
```

you have removed the dependency from the backtrace library.  All that is necessary now is to adjust the Spack recipe, which I can do.

We don't have documentation for writing CMake files within Phlex—that will likely become clearer once @greenc-FNAL has migrated Phlex from bare CMake to Cetmodules.

knoepfel (approved):

wddgit:
Thanks for reviewing and merging. Is there a command or recipe for running clang-format on a working area?

knoepfel:
@wddgit, the goal is to integrate `clang-format` as part of the PR process so we don't have to do this by hand.  But for now, one option is to go to the top-level directory of the repository and type:

```console
> clang-format -i -style=file $(git diff HEAD^ --name-only | grep "\.cpp\|\.hpp")
```

This will apply clang-format to the `.cpp` and `.hpp` files that have been changed since the previous commit.  

One downside to this approach is that the settings in Phlex's `.clang-format` file aren't supported by each version of `clang-format` that one of us might use.  I had to tweak some things by hand to get clang-format to work on `scisoftbuild02`, and I didn't push those changes to the repository.  This issue should be largely fixed when integrating clang-format as part of the PR.

knoepfel:
Yes, the CMS bot system has been effective for doing this type of thing—LArSoft also uses it.  It was created, however, before GitHub supported "actions", which are natively supported by GitHub and designed for doing these kinds of checks.

For our own development, I think Phlex should make use of GitHub's facilities until it's clear that they do not meet our needs.  This insulates us from depending on a home-grown solution (which requires maintenance) for this particular need.

PR #22, opened 2025-08-06
Use configuration for I/O, plus general cleanup

barnaliy:
This PR reorganizes the configuration system that maps data products to storage technologies and file destinations.

What the Configuration System Does:
The configuration system allows users to specify:

Which storage technology to use for each data product (ROOT TTree, ROOT RNtuple, HDF5)
Which file each data product should be written to
Runtime flexibility rather than compile-time hardcoded storage decisions.

Tagging @gemmeren and @aolivier23 
Requesting reviews from @knoepfel  and @pcanal


form/form/form.cpp
```diff
@@ -3,53 +3,83 @@
...
+  // Look up creator from config
   void form_interface::write(const std::string& creator,
-                             const std::vector<phlex::product_base>& batch)
+                             const std::vector<phlex::testing::product_base>& batch)
   {
     if (batch.empty())
       return;
 
+    // Look up creator from config based on product name.
+    auto it = m_product_to_config.find(batch[0].label);
```
aolivier23, line 49: Is this an unused variable?  The comment makes it sound like this is from a previous draft of form.cpp that didn't make it into the PR.

form/form/form.cpp
```diff
@@ -3,53 +3,83 @@
...
     const std::string& id = batch[0].id;
     m_pers->commitOutput(creator, id);
   }
 
-  void form_interface::read(phlex::product_base& pb)
+  void form_interface::read(const std::string& creator, phlex::testing::product_base& pb)
   {
-    // For now reading into 'empty' product_base, may change
+    // Look up creator from config based on product name. O(1) lookup instead of loop
+    auto it = m_product_to_config.find(pb.label);
```
aolivier23, line 74: Unused variable from previous draft?

gemmeren, line 74: As @aolivier23 indicated above, these section may not be needed.

test/form/config.cpp
```diff
@@ -0,0 +1,22 @@
...
+namespace phlex::config {
+
+  void parse_config::addItem(const std::string& product_name,
+                             const std::string& file_name,
+                             int technology)
+  {
+    m_items.emplace_back(product_name, file_name, technology);
+  }
+
+  const PersistenceItem* parse_config::findItem(const std::string& product_name) const
```
aolivier23, line 12: Could we return `std::optional<>` here instead of a raw pointer like you did in `phlex/parse_config.cpp`?

test/form/data_products/track_start.cpp
```diff
@@ -10,6 +10,12 @@ float TrackStart::getY() const { return m_y; }
 
 float TrackStart::getZ() const { return m_z; }
 
+void TrackStart::setX(float x) { m_x = x; }
```
aolivier23, line 13: Didn't we remove this in an earlier version of FORM to satisfy a PHLEX PR comment?

test/form/data_products/track_start.hpp
```diff
@@ -16,6 +16,10 @@ class TrackStart {
   float getY() const;
   float getZ() const;
 
+  void setX(float x);
```
aolivier23, line 19: Can we remove these?  I think we removed them in a previous version of FORM to satisfy a PHLEX PR comment.

form/persistence/persistence.cpp
```diff
@@ -9,71 +9,143 @@ using namespace form::detail::experimental;
...
+      }
+    }
+  }
+  if (!found) {
+    throw std::runtime_error("No configuration found for product: " + product_name +
+                             " from creator: " + creator);
+  }
+
+  std::string full_label;
+  if (product_name == "/index") {
```
gemmeren, line 106: Is the if clause needed?
why not:
`full_label = creator + "/" + product_name;`
unconditionally?

aolivier23 (commented):
Please check on the possible regression in `track_start.hpp` and .cpp.  Should be an easy fix!

Please also consider making the interface to `phlex::parse_config` look more like `form::experimental::config::parse_config`. Returning a raw pointer isn't actively causing problems.  But it does leave us open to "where did that come from?!" bugs down the road.

form/persistence/persistence.cpp
barnaliy: If you did full_label = creator + "/" + product_name unconditionally:

For regular products: creator + "/" + "trackStart" = "creator/trackStart" 
For index: creator + "/" + "/index" = "creator//index" ✗ (double slash)

So yes, the index DOES come with a forward slash - it's hardcoded as "/index" in the calls to getPlacement(). if clause id needed.

gemmeren: Yeah, so kind of wondering whether the '/' in product name for index is really a good idea...

form/persistence/persistence.cpp
```diff
@@ -9,71 +9,143 @@ using namespace form::detail::experimental;
...
-  std::unique_ptr<Placement> plcmnt = getPlacement(label);
+  std::unique_ptr<Placement> plcmnt = getPlacement(creator, label);
   m_store->fillContainer(*plcmnt, data, type);
   return;
 }
 
 void Persistence::commitOutput(const std::string& creator, const std::string& id)
 {
-  std::unique_ptr<Placement> plcmnt = getPlacement(creator + "/index");
+  std::unique_ptr<Placement> plcmnt = getPlacement(creator, "/index");
```
gemmeren, line 49: You could have decided here not to include the '/', since its no longer needed as separator ;)

form/persistence/persistence.cpp
```diff
@@ -9,71 +9,143 @@ using namespace form::detail::experimental;
...
+std::unique_ptr<Placement> Persistence::getPlacement(const std::string& creator,
+                                                     const std::string& label)
 {
-  // FIXME: get parameters from configuration
-  std::unique_ptr<Placement> plcmnt = std::unique_ptr<Placement>(new Placement("toy.root", label, 1 * 256 + 1));
+  //use config to determine values
+  std::string file_name;
+  int technology;
+
+  std::string product_name = label;
```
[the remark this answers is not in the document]
barnaliy: Yes, I agree - this is an unnecessary string copy.
Looking at how product_name is used in the getPlacement function:

It's only used for comparisons: if (item.product_name == product_name)
It's used in error messages: "No configuration found for product: " + product_name
Since product_name is never modified after the assignment, we could replace it with:
const std::string& product_name = label;  // Reference instead of copy.

Thanks for pointing this out.

form/persistence/persistence.cpp
gemmeren: Thanks ;)

test/form/config.cpp
barnaliy: Thanks! Changed to std::optional for better consistency with phlex/parse_config.cpp.

aolivier23 (approved):
Thank you for fixing my comments!

form/form/technology.hpp
```diff
@@ -0,0 +1,15 @@
+#ifndef __TECHNOLOGY_HPP__
```
gemmeren, line 1: Please be consistent with the "_"'s

form/persistence/persistence.cpp
```diff
@@ -9,71 +9,133 @@ using namespace form::detail::experimental;
...
 {
-  // FIXME: get parameters from configuration
-  std::unique_ptr<Placement> plcmnt = std::unique_ptr<Placement>(new Placement("toy.root", label, 1 * 256 + 1));
+  //use config to determine values
+  std::string file_name;
+  int technology;
+
+  // Special handling for index containers
+  bool found = false;
+  if (label == "index") {
```
gemmeren, line 76: is this if clause really needed?

form/persistence/persistence.cpp
```diff
@@ -9,71 +9,133 @@ using namespace form::detail::experimental;
...
-    index = label.substr(0, del_pos) + "/index";
-  } else {
-    index = label + "/index";
+  // Full label and index label construction
+  std::string full_label = creator + "/" + label;
+  std::string index_label = creator + "/index";
+
+  // Get parameters from configuration
+  std::string file_name;
+  int technology;
```
gemmeren, line 119: You have this fragment a few times, maybe add a m_config.find(label, file&, tech&)?

form/form/CMakeLists.txt
```diff
@@ -10,6 +10,7 @@ add_library(form
       )
 
 # Set up include directories so form.hpp can find mock_phlex headers
+include_directories(${PROJECT_SOURCE_DIR})
```
gemmeren, line 13: Hmm, not exactly sure why this [w|sh]ould be needed, can you explain?

form/form/form.hpp
```diff
@@ -15,7 +15,7 @@ namespace form::experimental {
   class form_interface {
   public:
     form_interface(std::shared_ptr<phlex::testing::product_type_names> tm,
-                   const form::experimental::config::parse_config& config);
+                   const phlex::config::parse_config& phlex_config); // Accept phlex config
```
gemmeren, line 18: Didn't we agree on a "phlex::testing..." namespace to indicate this is _not_ real phlex code, but only used to test some constructs?

form/form/technology.hpp
gemmeren: I looked at the code again and believe before this PR, we/form consistently includes "__" in the #ifndef's and I don't see a reason to change that. So, I suggest leaving all '_' in unless phlex has a policy otherwise.

form/mock_phlex/phlex_toy_core.hpp
```diff
@@ -1,15 +1,15 @@
...
+#define PHLEX_TOY_CORE_HPP
 
 #include <memory>
 #include <string>
 #include <typeindex>
 #include <unordered_map>
 #include <vector>
 
-namespace phlex {
+namespace phlex::testing {
```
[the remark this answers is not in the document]
gemmeren: Just a note, I agree with namespace "mock_phlex", as long as it is used consistently, probably prefer dropping ::testing, since that aspect is included.

form/form/parse_config.hpp
```diff
@@ -0,0 +1,46 @@
+#ifndef FORM_PARSE_CONFIG_HPP
+#define FORM_PARSE_CONFIG_HPP
+
+#include "technology.hpp" // Use shared Technology constants
```
gemmeren, line 4: Where do you really need/use those?

gemmeren:
Why does technology _have_ to be part of config? it can (or should) be a more general construct of form, since you are also using it in other places, like factories. So, I would go with none of your solutions.

form/persistence/persistence.cpp
```diff
@@ -9,71 +9,133 @@ using namespace form::detail::experimental;
...
+        break;
+      }
+    }
+  }
+  if (!found) {
+    throw std::runtime_error("No configuration found for product: " + label +
+                             " from creator: " + creator);
+  }
+
+  std::string full_label = creator + "/" + label;
```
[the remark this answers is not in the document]
barnaliy: Addressed in the recent commit. thanks!

form/persistence/persistence.cpp
barnaliy: Addressed - Added findConfigItem() helper function to eliminate code duplication between getPlacement() and getToken().

test/form/config.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+#include <memory>
+#include <optional>
+#include <string>
+#include <vector>
+
+namespace phlex::config {
+
+  // Technology constants
+  namespace Technology {
```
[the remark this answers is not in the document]
gemmeren: but why do we even have these files here?

test/form/config.hpp
gemmeren, line 1: remove?

test/form/reader.cpp
```diff
@@ -1,8 +1,9 @@
 // Copyright (C) 2025 ...
 
+#include "config.hpp"
```
gemmeren, line 3: ditto

form/form/parse_config.cpp
```diff
@@ -0,0 +1,22 @@
...
+  {
+    for (const auto& item : m_items) {
+      if (item.product_name == product_name) {
+        return item;
+      }
+    }
+    return std::nullopt;
+  }
+
+} // namespace form::config
```
gemmeren, line 22: // namespace form::experimental::config

form/form/parse_config.hpp
```diff
@@ -0,0 +1,44 @@
...
+    std::optional<PersistenceItem> findItem(const std::string& product_name) const;
+
+    // Get all items (for debugging/validation)
+    const std::vector<PersistenceItem>& getItems() const { return m_items; }
+
+  private:
+    std::vector<PersistenceItem> m_items;
+  };
+
+} // namespace form::config
```
gemmeren, line 42: // namespace form::experimental::config

form/mock_phlex/phlex_toy_config.cpp
```diff
@@ -0,0 +1,22 @@
...
+  {
+    for (const auto& item : m_items) {
+      if (item.product_name == product_name) {
+        return &item;
+      }
+    }
+    return nullptr;
+  }
+
+} // namespace phlex::config
```
gemmeren, line 22: // namespace mock_phlex::config

form/mock_phlex/phlex_toy_config.hpp
```diff
@@ -0,0 +1,41 @@
...
+    const PersistenceItem* findItem(const std::string& product_name) const;
+
+    // Get all items (for debugging/validation)
+    const std::vector<PersistenceItem>& getItems() const { return m_items; }
+
+  private:
+    std::vector<PersistenceItem> m_items;
+  };
+
+} // namespace phlex::config
```
gemmeren, line 39: // namespace mock_phlex::config

form/util/factories.hpp
gemmeren, line 1: here you could include technology.hpp and replace the 'magic' numbers.

gemmeren (commented):
Mostly minor things now :)

form/mock_phlex/phlex_toy_core.hpp
```diff
@@ -1,15 +1,15 @@
...
+#define __PHLEX_TOY_CORE_HPP__
 
 #include <memory>
 #include <string>
 #include <typeindex>
 #include <unordered_map>
 #include <vector>
 
-namespace phlex {
+namespace mock_phlex::testing {
```
pcanal, line 12:
```suggestion
namespace mock_phlex {
```

form/util/factories.hpp
```diff
@@ -41,15 +43,16 @@ namespace form::detail::experimental {
 
   std::shared_ptr<IStorage_Container> createContainer(int tech, const std::string& name)
   {
-    if (int(tech / 256) == 1) {   //ROOT major technology
-      if (int(tech % 256) == 1) { //ROOT TTree minor technology
+    if (int(tech / 256) == form::Technology::ROOT_MAJOR) {
+      if (int(tech % 256) == form::Technology::ROOT_TTREE_MINOR) {
```
pcanal, line 47:
```suggestion
    if (form::Technology::GetMajor(tech) == form::Technology::ROOT_MAJOR) {
      if (form::Technology::GetMinor(tech) == form::Technology::ROOT_TTREE_MINOR) {
```
or
```suggestion
    if (form::Technology::MatchMajor(tech, form::Technology::ROOT_MAJOR)) {
      if (form::Technology::MatchMinor(tech, form::Technology::ROOT_TTREE_MINOR)) {
```

form/form/technology.hpp
```diff
@@ -0,0 +1,25 @@
...
+#define __TECHNOLOGY_HPP__
+
+namespace form {
+
+  // Technology constants - single source of truth
+  namespace Technology {
+    const int ROOT_TTREE = 1 * 256 + 1;   // ROOT TTree
```
pcanal, line 8:
```suggestion
    constexpr int Combine(int major, int minor) { return major * 256 + minor; };
    
    const int ROOT_TTREE = Combine(ROOT_MAJOR, ROOT_TREE_MINOR);
```

form/root_storage/root_tfile.cpp
```diff
@@ -15,8 +15,4 @@ ROOT_TFileImp::ROOT_TFileImp(const std::string& name, char mode) :
   }
 }
 
-//Trivial destructor is sufficient because all members managed by the stack.
-//Writing it here lets us move the dependence on the TFile implementation out of the header.
-ROOT_TFileImp::~ROOT_TFileImp() {}
```
knoepfel, line 20: Why move this back in the header file?  If you want to use the default-generated destructor, you can do that here in the `.cpp` file:

```cpp
ROOT_TFileImp::~ROOT_TFileImp() = default;
```

form/storage/storage_associative_container.hpp
```diff
@@ -14,10 +14,10 @@ namespace form::detail::experimental {
     Storage_Associative_Container(const std::string& name);
     ~Storage_Associative_Container() = default;
 
-    const std::string& top_name() { return m_tName; }
-    const std::string& col_name() { return m_cName; }
+    const std::string& top_name();
+    const std::string& col_name();
```
knoepfel, line 18: Should be `const`-qualified member functions:

```suggestion
    const std::string& top_name() const;
    const std::string& col_name() const;
```

form/storage/storage_container.cpp
```diff
@@ -7,8 +7,6 @@ using namespace form::detail::experimental;
 
 Storage_Container::Storage_Container(const std::string& name) : m_name(name), m_file(nullptr) {}
 
-Storage_Container::~Storage_Container() {}
```
knoepfel, line 10: Same comment as before.  Putting the implementation of the destructor in the `.cpp` makes sense.  If you want it to be the default implementation, just specify it here:

```cpp
Storage_Container::~Storage_Container() = default;
```

form/persistence/persistence.cpp
```diff
@@ -2,78 +2,123 @@
...
 using namespace form::detail::experimental;
 
 // Factory function implementation
 namespace form::detail::experimental {
-  std::unique_ptr<IPersistence> createPersistence()
+  std::unique_ptr<IPersistence> createPersistence(
+    const form::experimental::config::parse_config& config) //factory takes form config
   {
-    return std::unique_ptr<IPersistence>(new Persistence());
+    return std::unique_ptr<IPersistence>(new Persistence(config));
```
knoepfel, line 16: If you're adjusting this line of code anyway, I suggest using `std::make_unique:`

```suggestion
    return std::make_unique<Persistence>(config);
```

form/persistence/persistence.cpp
```diff
@@ -2,78 +2,123 @@
...
   }
 } // namespace form::detail::experimental
 
-Persistence::Persistence() : m_store(nullptr) { m_store = createStorage(); }
+Persistence::Persistence(const form::experimental::config::parse_config& config) :
+  m_store(nullptr), m_config(config) // constructor takes form config
+{
+  m_store = createStorage();
+}
```
knoepfel, line 24:
```suggestion
  m_store(createStorage()), m_config(config) // constructor takes form config
{}
```

PR #23, opened 2025-08-12
Ensure that ReadVector runs after WriteVector

knoepfel:
I'm encountering occasional test failures whenever the `ReadVector` CTest executes (something about `toy.root` not existing).  I *think* this is because the `ReadVector` test requires `toy.root` to have first been produced by the `WriteVector` test.  When invoking`ctest -j<N>`, this ordering is not guaranteed unless the dependency between the two tests is expressed.  This PR adds that dependency.


gemmeren (approved):

aolivier23 (approved):
Good catch!

PR #26, opened 2025-08-19
Apply clang-format 20

knoepfel:
This PR: 

1. Applies clang-format 20 to the code base, using the C++ coding style [here](https://github.com/Framework-R-D/phlex-coding-guidelines/blob/main/CppGuidelines.md)
2. Creates a `format-check` GitHub action that checks for conformance with the coding style
3. Creates a `format-fix` GitHub action that allows users to specify `@phlexbot format`, which will then execute `clang-format` on the PR and commit any changes as a separate commit to the PR.


aolivier23 (approved):
Looks good to me, and I think I understand what the new CI action does.  Thank you for getting this formatting done.  I'll remember to updated FORM's .clang-format when we're ready.

aolivier23:
Sounds brilliant to me.  May I port it over to the FORM development area?

.github/workflows/format-fix.yaml
```diff
@@ -0,0 +1,75 @@
...
+          source: "./phlex-src"
+          clangFormatVersion: 20
+          inplace: True
+
+      - uses: EndBug/add-and-commit@v9
+        id: add_and_commit
+        with:
+          cwd: phlex-src
+          author_name: "github-actions[bot]"
+          author_email: "41898282+github-actions[bot]@users.noreply.github.com"
```
greenc-FNAL, line 60: Should this be generated somehow or filled in via `${{ ... }}`?

greenc-FNAL (commented):
A couple of minor questions:

1. Have you configured `dependabot` to operate on these repositories so that dependencies on external actions are updated automatically (or at least, via automated PR)?
2. I'm a little concerned that most of the action dependencies are apparently third-party rather than from GitHub proper or a similarly weighty source, but perhaps I'm just ignorant of the ecosystem.
3. Can you remind me whether `clang-format` will honor superseding `.clang-format` files as it descends the directory tree?

knoepfel:
@greenc-FNAL, thanks for your comments:

> 1. Have you configured dependabot to operate on these repositories so that dependencies on external actions are updated automatically (or at least, via automated PR)?

No, I have not.  That can be a future enhancement.

> 2. I'm a little concerned that most of the action dependencies are apparently third-party rather than from GitHub proper or a similarly weighty source, but perhaps I'm just ignorant of the ecosystem.

The marketplace shares community-developed solutions.  Some of these solutions may go away over time, but the ones chosen here do the right thing, and we don't have to develop them ourselves.  There are relatively few actions that are directly supported by GitHub proper.

> 3. Can you remind me whether clang-format will honor superseding .clang-format files as it descends the directory tree?

Yes, it will.

.github/workflows/format-fix.yaml
greenc-FNAL: Not necessary, I just worried it was somehow meaningful to GitHub in a way that affected record keeping or traceability.

greenc-FNAL:
> > 3\. Can you remind me whether clang-format will honor superseding .clang-format files as it descends the directory tree?
> 
> Yes, it will.

Great, that means that the FORM coding style can be controlled by them and doesn't have to match that of Phlex proper just because that's where it is right now.

Issue #29, opened 2025-08-22
Too many CMake warnings about `dot`


pcanal:
Running `CMake` for `Phlex` without `dot` lead to many warning like:
```
CMake Warning at test/CMakeLists.txt:10 (message):
  dot program not found; all TEST_DOT_GRAPH options will be ignored.
Call Stack (most recent call first):
  test/CMakeLists.txt:55 (dot_test)
  test/CMakeLists.txt:64 (add_catch_test)
```
independently from whether dot is needed or whether the warning is useless, we really only need the warning once.

knoepfel:
Resolved by #30.

PR #30, opened 2025-08-22
Warn only once per build directory about missing DOT

pcanal:
This mitigates #29  ...

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1,19 +1,17 @@
 find_package(Catch2 REQUIRED)
 
 find_program(DOT dot)
+if (NOT DOT)
+  message (
+    WARNING
+    "dot program not found; all TEST_DOT_GRAPH options will be ignored."
+  )
+endif()
 
 add_definitions("-UNDEBUG")
 
 function (dot_test BASENAME)
-  if (NOT DOT)
-    if (NOT DEFINED DONE_DOT_WARNING)
-      message(
-        WARNING
-          "dot program not found; all TEST_DOT_GRAPH options will be ignored."
-        )
-      set(DONE_DOT_WARNING TRUE CACHE BOOL "suppress silly warning")
-    endif()
-  else ()
+  if (DOT)
     set(ORIG_TEST_DIR ${CMAKE_CURRENT_BINARY_DIR}/${BASENAME}.d)
     set(TEST_DIR ${CMAKE_CURRENT_BINARY_DIR}/dot_${BASENAME}.d)
     file(MAKE_DIRECTORY ${TEST_DIR})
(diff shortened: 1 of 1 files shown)


knoepfel (commented):
Instead of making a cache variable, I think I'd just do something like:

```diff
  find_program(DOT dot)  # Line 3 of file
+ if (NOT DOT)
+   message(
+     WARNING
+        "dot program not found; all TEST_DOT_GRAPH options will be ignored."
+   )
+ endif ()
```

then simplify the `dot_test` function:

```cmake
function (dot_test BASENAME)
  set(ORIG_TEST_DIR ${CMAKE_CURRENT_BINARY_DIR}/${BASENAME}.d)
  set(TEST_DIR ${CMAKE_CURRENT_BINARY_DIR}/dot_${BASENAME}.d)
  ...
endfunction ()
```

and then make the change:

```diff
- if (UT_TEST_DOT_GRAPH)
+ if (UT_TEST_DOT_GRAPH AND DOT)
      dot_test(${BASENAME})
    endif ()
```

pcanal:
> and then make the change:

There is a second place that call `dot_test`.  Do prefer to put the `if` statement in both location or simply inside the function?

> Instead of making a cache variable, I think I'd just do something like:

With the cache variable (whether it is inside the function or not), the message is printed once per build directory (on the first invocation of CMake).  Without the change variable, it is there for each invocation of CMake. 

If `dot` is really minor, I would prefer to either print it once per directory or to change from a warning to an information message:
```
cmake .
CMake Warning at test/CMakeLists.txt:5 (message):
  dot program not found; all TEST_DOT_GRAPH options will be ignored.


-- Configuring done (11.9s)
-- Generating done (1.9s)
-- Build files have been written to: /home/pcanal/phlex_working/build/phlex-standalone-cppyy
```
vs
```
cmake .
-- dot program not found; all TEST_DOT_GRAPH options will be ignored.
-- Configuring done (0.9s)
-- Generating done (0.2s)
-- Build files have been written to: /home/pcanal/phlex_working/build/phlex-standalone-cppyy
```
vs
```
cmake .
-- Configuring done (0.9s)
-- Generating done (0.2s)
-- Build files have been written to: /home/pcanal/phlex_working/build/phlex-standalone-cppyy
```

pcanal:
I went for the informational message, the current PR code issues:
```
cmake .
-- dot program not found; all TEST_DOT_GRAPH options will be ignored.
-- Configuring done (0.9s)
-- Generating done (0.2s)
-- Build files have been written to: /home/pcanal/phlex_working/build/phlex-standalone-cppyy
```

knoepfel (approved):
I like this, @pcanal.  Thanks

PR #36, opened 2025-09-05
FORM container config

aolivier23:
Adding container and file configuration attributes to FORM along with a few smaller recent changes.  File configuration attributes let us configure the compression algorithm TFile uses for example without a TFile-specific class in the interface.  This will also eventually be important for RNTuple.

### DIFF
--- a/form/persistence/persistence.cpp
+++ b/form/persistence/persistence.cpp
@@ -75,7 +75,11 @@ form::experimental::config::PersistenceItem const* Persistence::findConfigItem(
   std::string const& label) const
 {
   auto const& items = m_output_items.getItems();
-  if(label == "index") return (items.empty()) ? nullptr : &(*items.begin());  //emulate how FORM did this before Phlex PR #22.  Will be fixed in a future FORM update.
+  if (label == "index")
+    return (items.empty())
+             ? nullptr
+             : &(*items
+                    .begin()); //emulate how FORM did this before Phlex PR #22.  Will be fixed in a future FORM update.
 
   auto it = std::find_if(
     items.begin(), items.end(), [&label](auto const& item) { return item.product_name == label; });
@@ -121,6 +125,8 @@ std::unique_ptr<Token> Persistence::getToken(std::string const& creator,
   std::string const full_label = buildFullLabel(creator, label);
   std::string const index_label = buildFullLabel(creator, "index");
 
-  int const rowId = m_store->getIndex(Token{config_item->file_name, index_label, config_item->technology}, id, m_tech_settings);
-  return std::make_unique<Token>(config_item->file_name, full_label, config_item->technology, rowId);
+  int const rowId = m_store->getIndex(
+    Token{config_item->file_name, index_label, config_item->technology}, id, m_tech_settings);
+  return std::make_unique<Token>(
+    config_item->file_name, full_label, config_item->technology, rowId);
 }


aolivier23:
I checked that the build and format check work by hand after using phlexbot to fix a clang-format problem.

form/util/factories.hpp
```diff
@@ -21,37 +23,52 @@ namespace form::detail::experimental {
...
+    // Use the helper functions from Technology namespace for consistency
+    if (form::technology::GetMajor(tech) == form::technology::ROOT_MAJOR) {
+      if (form::technology::GetMinor(tech) == form::technology::ROOT_TTREE_MINOR) {
 #ifdef USE_ROOT_STORAGE
         return std::make_shared<ROOT_TBranch_ContainerImp>(name);
-#endif
+#endif // USE_ROOT_STORAGE
       }
+    } else if (form::technology::GetMajor(tech) == form::technology::HDF5_MAJOR) {
+#ifdef USE_HDF5
```
gemmeren, line 65: USE_HDF5_STORAGE  ?

form/form/form.hpp
```diff
@@ -26,7 +26,9 @@ namespace form::experimental {
   private:
     std::unique_ptr<form::detail::experimental::IPersistence> m_pers;
     std::shared_ptr<mock_phlex::product_type_names> m_type_map;
-    form::experimental::config::parse_config m_config;
+    form::experimental::config::output_item_config m_output_items;
```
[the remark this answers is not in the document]
aolivier23: You appear to be correct.  I'll confirm that everything still works then send these over in a commit.  Thank you for spotting this!

form/form/config.cpp
```diff
@@ -0,0 +1,47 @@
+#include "form/config.hpp"
+
+namespace {
+  template <class MAP_LIKE>
+  auto const_lookup(const MAP_LIKE& map, typename MAP_LIKE::key_type const& key)
```
knoepfel, line 5: This returns a copy.  Do you intend for this to be the case?

form/persistence/persistence.cpp
```diff
@@ -60,64 +66,69 @@ void Persistence::read(std::string const& creator,
...
-  int rowId = m_store->getIndex(*index_token, id);
-
-  return std::make_unique<Token>(
-    config_item->file_name, full_label, config_item->technology, rowId);
+  std::unique_ptr<Token> index_token =
+    std::unique_ptr<Token>(new Token(file_name, index_label, technology));
+  int rowId = m_store->getIndex(*index_token, id, m_tech_settings);
+  std::unique_ptr<Token> token =
+    std::unique_ptr<Token>(new Token(file_name, full_label, technology, rowId));
+  return token;
```
knoepfel, line 133: I believe this can be simplified:

```c++
int rowID = m_store->getIndex(Token{file_name, index_label, technology},
                              id,
                              m_tech_settings);
return std::make_unique<Token>(file_name, full_label, technology, rowId);
```

knoepfel (commented):
@aolivier23, thanks for the PR.  I have a few comments below.

My concern at this point is whether the user's overall Phlex-job configuration (nested tables, each of which apply to a specific entity) will map cleanly to FORM.  For example, it's possible in _art_ to specify different output streams (perhaps with different backend technologies), where the configuration options are specific to each output stream.  I don't see anything yet in FORM that prevents this, but I start to get nervous when I see virtual-function overrides that only throw exceptions.  That may indicate that a virtual function is not the best choice and another solution may be preferable.

For now, I suggest moving forward with what you have (modulo the comments below).  Once FORM adopts the configuration utilities provided by Phlex (to be fair, they're not fully fleshed out), we should take a look at whether the design should be adjusted.

form/persistence/persistence.cpp
aolivier23: Will do.

form/form/config.cpp
aolivier23: Will fix.  I do not intend to return a copy.  I think I can return a const reference.

form/persistence/persistence.cpp
```diff
@@ -60,64 +66,69 @@ void Persistence::read(std::string const& creator,
...
+  for (auto const& item : m_output_items.getItems()) {
+    // Special handling for index containers, take first file, tech, FIXME: Reconsider
+    if (item.product_name == label || label == "index") {
+      file_name = item.file_name;
+      technology = item.technology;
+      found = true;
+      break;
+    }
+  }
+  if (!found) {
```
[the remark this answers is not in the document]
aolivier23: Will do.  I plan to try `std::find_if()` first but am happy to go back to the version with `break` if `find_if()` makes things worse.

form/persistence/persistence.cpp
```diff
@@ -60,64 +66,69 @@ void Persistence::read(std::string const& creator,
...
                              " from creator: " + creator);
   }
 
-  std::string const full_label = buildFullLabel(creator, label);
-
-  return std::make_unique<Placement>(config_item->file_name, full_label, config_item->technology);
+  std::string full_label = creator + "/" + label;
+  std::unique_ptr<Placement> plcmnt =
+    std::unique_ptr<Placement>(new Placement(file_name, full_label, technology));
+  return plcmnt;
```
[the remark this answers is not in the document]
aolivier23: Will do.

form/persistence/persistence.cpp
```diff
@@ -60,64 +66,69 @@ void Persistence::read(std::string const& creator,
...
+    if (item.product_name == label) {
+      file_name = item.file_name;
+      technology = item.technology;
+      found = true;
+      break;
+    }
+  }
 
-  if (!config_item) {
+  if (!found) {
```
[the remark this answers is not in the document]
aolivier23: Will do.  Trying `std::find_if()` first.

aolivier23:
@phlexbot format

knoepfel (approved):

PR #38, opened 2025-09-17
Use Boost PFR in `node_catalog`

knoepfel:
Using Boost's PFR library, it's possible to retrieve a particular field from a `struct` based on the provided type.


wddgit (approved):
One question (not a suggestion one way or the other). Do you intend to remove the aliases like ```declared_predicates``` everywhere eventually?

Otherwise this looks good. It does bring in a dependence on boost PFR. I tried copilot on the PR and it helped me understand the PR quicker. It was able to explain what PFR was as I had never heard of it before.

PR #40, opened 2025-09-18
Verify that a data product can be a vector of abstract types

knoepfel:
In yesterday's Phlex development meeting, the question arose whether we should support data products with abstract types.  This PR demonstrates that the current implementation of Phlex already supports this.

All Phlex cares about is that the type of a produced product can be matched with the type of a consumed product.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -62,6 +62,7 @@ add_unit_test(yielding_driver LIBRARIES phlex::core TBB::tbb)
 add_catch_test(
   allowed_families LIBRARIES phlex::core Boost::json TEST_DOT_GRAPH
   )
+add_catch_test(vector_of_abstract_types LIBRARIES phlex::core)
 add_catch_test(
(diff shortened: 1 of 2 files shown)


greenc-FNAL:
Is this sufficient to answer the question re vector of `shared_ptr<abstract>`, or is that a different problem?

knoepfel:
> Is this sufficient to answer the question re vector of `shared_ptr<abstract>`, or is that a different problem?

Yes, to first order.  For a created data product of type `shared_ptr<abstract>`, any downstream node must consume that data products as the type `shared_ptr<abstract>`.  IOW, there is no automatic translation yet from `shared_ptr<abstract>` to `shared_ptr<derived>` (and vice versa).

However, for our discussion we had yesterday, it is fine to do something like:

```c++
shared_ptr<abstract> make_derived(some_type t) 
{ 
  // Transform from some_type to some derived class
  return std::make_shared<derived>(t); 
}
```

greenc-FNAL (approved):

PR #42, opened 2025-09-22
Apply clang-format to concept definitions

knoepfel:
Previous versions of `clang-format` did not understand the `concept` syntax.  The version of `clang-format` that we are using does, and this PR updates the relevant code accordingly.

### DIFF
--- a/phlex/core/concepts.hpp
+++ b/phlex/core/concepts.hpp
@@ -9,14 +9,14 @@
 #include <utility>
 
 namespace phlex::experimental {
-  // clang-format off
-  //  => clang-format 15 does support concept declarations well
+
   template <typename T>
   concept not_void = !std::same_as<T, void>;
 
   template <typename T>
-  concept sendable = std::move_constructible<T> ||
-    requires(T& t) { {send(t)} -> std::move_constructible; };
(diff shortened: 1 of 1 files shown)


phlex/core/concepts.hpp
```diff
@@ -55,13 +54,11 @@ namespace phlex::experimental {
...
   concept is_output_like = std::is_member_function_pointer_v<T> &&
-                           expects_input_parameters<T, product_store const&> &&
-                           returns<T, void>;
+                           expects_input_parameters<T, product_store const&> && returns<T, void>;
 
   template <typename T>
   concept is_fold_like =
-    at_least_two_input_parameters<T> &&
-    first_input_parameter_is_non_const_lvalue_reference<T> &&
+    at_least_two_input_parameters<T> && first_input_parameter_is_non_const_lvalue_reference<T> &&
```
greenc-FNAL, line 61: Is there a way to require that concepts are formatted with only one logical clause per line, and is this something we should do?

greenc-FNAL (changes requested):

greenc-FNAL (approved):

PR #45, opened 2025-09-23
Python support using converter nodes

wlav:
This PR adds Python support through input/output converter nodes to solve the combinatorical problem. This approach makes all template instantiations compile-time only, thus cutting down on startup and memory overhead compared to using a JIT. They do add a per-algorithm run-time overhead.

Current support is for builtin types and `std::vector` of builtin types. Full support of all types requires some code generation from the eventual IDL. Life-times are managed using "lifelines", tying input underlying to output views, as Phlex can not track through converter nodes.

This PR adds support for registration that mimics C++ in naming, while remaining Pythonic and using reflection to cut down on boilerplate code where possible. It also adds a basic Pythonic representation of the configuration, with the actual remaining in C++. Error reporting translates Python tracebacks into C++ `std::runtime_error` exceptions, for handling and printing bye the C++ main.

The main open question is on interpreter finalization, which this PR punts on. This awaits resolution for when Phlex supports tokens.

This PR is ready to merge: the final failing checks are on code coverage of error reporting paths. These are in fact exercised: the tool is wrong.

### DIFF
--- a/phlex/app/phlex.cpp
+++ b/phlex/app/phlex.cpp
@@ -92,5 +92,13 @@ int main(int argc, char* argv[])
   if (not vm["parallel"].defaulted()) {
     max_concurrency = vm["parallel"].as<int>();
   }
-  phlex::experimental::run(configurations, max_concurrency);
+  try {
+    phlex::experimental::run(configurations, max_concurrency);
+  } catch (std::exception const& e) {
+    std::cerr << e.what() << '\n';
+    return 1;
+  } catch (...) {
+    std::cerr << "Unknown exception caught.\n";
+    return 1;
+  }
 }


plugins/python/pymodule.cpp
```diff
@@ -0,0 +1,134 @@
...
+  initialize();
+
+  PyGILRAII g;
+
+  std::string modname = config.get<std::string>("pyplugin");
+  PyObject* mod = PyImport_ImportModule(modname.c_str());
+  if (mod) {
+    PyObject* reg = PyObject_GetAttrString(mod, "PHLEX_EXPERIMENTAL_REGISTER_ALGORITHMS");
+    if (reg) {
+      PyObject* pym = wrap_module(&m);
```
beojan: `wrapped_module`?

plugins/python/pymodule.cpp
```diff
@@ -0,0 +1,134 @@
...
+  void* addr = dlsym(RTLD_DEFAULT, "Py_IsInitialized");
+  if (addr) {
+    Dl_info info;
+    if (dladdr(addr, &info) == 0 || info.dli_fname == 0 || info.dli_fname[0] == 0) {
+      throw std::runtime_error("unable to determine linked libpython");
+    }
+    dlopen(info.dli_fname, RTLD_GLOBAL | RTLD_NOW);
+  }
+
+#if PY_VERSION_HEX < 0x03020000
```
beojan: Is support for such old Python versions required?

plugins/python/pymodule.cpp
```diff
@@ -0,0 +1,134 @@
...
+  PyType_Ready(&PhlexConfig_Type);
+  PyType_Ready(&PhlexModule_Type);
+  PyType_Ready(&PhlexLifeline_Type);
+
+  // load numpy (see also above, if already initialized)
+#ifdef PHLEX_HAVE_NUMPY
+  import_numpy(true);
+#endif
+
+  // TODO: the GIL should first be released on the main thread and this seems
```
beojan: If you use a static RAII type that releases the GIL on construction and re-acquires on destruction that should solve this, and automatically handle only releasing it once:
https://en.cppreference.com/w/cpp/language/storage_duration.html#Static_block_variables

plugins/python/configwrap.cpp
```diff
@@ -0,0 +1,209 @@
...
+    if (!coll) {
+      try {
+        auto const& cvalue = pycmap->ph_config->get<long>(cname);
+        pyvalue = PyLong_FromLong(cvalue);
+      } catch (...) {
+        PyErr_Format(PyExc_TypeError, "property \"%s\" is not an integer", cname.c_str());
+      }
+    } else {
+      try {
+        auto const& cvalue = pycmap->ph_config->get<std::vector<std::string>>(cname);
```
beojan: Presumably the issue is elsewhere, but why is this read from the configuration as a list of strings and converted to a list of integers, instead of being read as a list of integers to begin with?

plugins/python/configwrap.cpp
```diff
@@ -0,0 +1,209 @@
...
+      }
+    }
+  } else if (type) {
+    PyErr_SetString(PyExc_TypeError, "requested type not supported");
+  }
+
+  if (type)
+    return pyvalue; // may be nullptr
+
+  // untyped (guess) conversion
```
beojan: Can't we introspect the value (using [this](https://www.boost.org/doc/libs/latest/libs/json/doc/html/ref/value/kind.html), or the `is_*` members) and just generate the correct type?

plugins/python/configwrap.cpp
```diff
@@ -0,0 +1,209 @@
...
+  }
+  if (!pyvalue) {
+    try {
+      auto const& cvalue = pycmap->ph_config->get<std::string>(cname);
+      pyvalue = PyUnicode_FromStringAndSize(cvalue.c_str(), cvalue.size());
+    } catch (std::runtime_error const&) {
+    }
+  }
+
+  // cache if found
```
beojan: Doesn't look like it

plugins/python/modulewrap.cpp
```diff
@@ -0,0 +1,632 @@
...
+
+  // insert input converter nodes into the graph
+  for (size_t i = 0; i < (size_t)cinputs.size(); ++i) {
+    // TODO: this seems overly verbose and inefficient, but the function needs
+    // to be properly types, so every option is made explicit
+    auto const& inp = cinputs[i];
+    auto const& inp_type = input_types[i];
+
+    if (inp_type == "bool")
+      INSERT_INPUT_CONVERTER(bool, inp);
```
beojan: If we have more than one Python algorithm using the same inputs, wouldn't this register the same conversion algorithm multiple times?

plugins/python/modulewrap.cpp
```diff
@@ -0,0 +1,632 @@
...
+
+  // register Python algorithm
+  if (cinputs.size() == 1) {
+    if (!output.empty()) {
+      auto* pyc = new py_callback_1{callable}; // TODO: leaks, but has program lifetime
+      mod->ph_module->transform(cname, *pyc, concurrency::serial)
+        .input_family(cinputs[0] + "py")
+        .output_products("py" + output);
+    } else {
+      auto* pyc = new py_callback_1v{callable}; // id.
```
beojan: It's not obvious that the `v` suffix here is for `void` and not `vector`.

beojan (commented):
I can't help but think this could be made a lot shorter and easier to read by using Pybind11 (which supports embedding the Python interpreter). That would be a fair amount of work though, and probably not worth it for the prototype.

plugins/python/configwrap.cpp
```diff
@@ -0,0 +1,209 @@
...
+        for (Py_ssize_t i = 0; i < (Py_ssize_t)cvalue.size(); ++i) {
+          PyObject* item = PyLong_FromString(cvalue[i].c_str(), nullptr, 10);
+          PyTuple_SetItem(pyvalue, i, item);
+        }
+      } catch (...) {
+        PyErr_Format(
+          PyExc_TypeError, "property \"%s\" is not a collection of integers", cname.c_str());
+      }
+    }
+  } else if (type) {
```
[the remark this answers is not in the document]
wlav: All allowable JSON types are not supported through the use of kind().

plugins/python/errorwrap.cpp
```diff
@@ -0,0 +1,43 @@
...
+  PyObject* exc = PyErr_GetRaisedException();
+  if (exc) {
+    PyObject* pymsg = PyObject_Str(exc);
+    msg = PyUnicode_AsString(pymsg);
+    Py_DECREF(pymsg);
+    Py_DECREF(exc);
+  }
+#endif
+
+  throw std::runtime_error(msg.c_str());
```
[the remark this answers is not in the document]
wlav: This is now implemented (and yes, it's _really_ nice).

plugins/python/modulewrap.cpp
wlav: There's a greater split in use now b/c of the existence of separate transform and observe functions; I'll continue to see whether there's a hope of further cleanup, but these callbacks are now no longer used in the same function.

plugins/python/src/configwrap.cpp
```diff
@@ -0,0 +1,209 @@
...
+  }
+  if (!pyvalue) {
+    try {
+      auto const& cvalue = pycmap->ph_config->get<std::string>(cname);
+      pyvalue = PyUnicode_FromStringAndSize(cvalue.c_str(), cvalue.size());
+    } catch (std::runtime_error const&) {
+    }
+  }
+
+  // cache if found
```
beojan: Marking resolved because I don't think this should be a blocker.

beojan (changes requested):
There's one change that I think is a blocker for merging this, which is that the new `kind` member on the configuration should have a name that makes it clear that it is not for external use.

greenc-FNAL (approved):
I have a WIP addressing some of the remaining test coverage issues, but I believe this PR is good to go as is.

PR #46, opened 2025-10-01
FORM RNTuple Support

aolivier23:
Introduces RNTuple as a file technology FORM can use to read and write to .root files.  RNTuple support is introduced through the `Storage_Container` interface, just like how TTree technology works.  Automatically adapts to classes that have custom streamers like `TLorentzVector`.


aolivier23:
@phlexbot format

PR #52, opened 2025-10-03
Copilot instructions

greenc-FNAL:


.github/copilot-instructions.md
```diff
@@ -0,0 +1,100 @@
...
+- If the developer's request contains an error or misunderstanding, explain the issue clearly
+
+### Truth and Accuracy
+
+Accuracy and honesty are critical:
+
+- If you lack sufficient information to complete a task, say so explicitly: "I don't know" or "I don't have access to the information needed"
+- Ask the developer for help or additional information when needed
+- Never fabricate answers or hide gaps in knowledge
+- It is better to acknowledge limitations than to provide incorrect information
```
pcanal, line 20: Does the order matter? 

(It probably does not matter but) Are all 4 requests necessary they fell somewhat overlapping (partially redundant)

.github/copilot-instructions.md
```diff
@@ -0,0 +1,100 @@
...
+### Clear and Direct Communication
+
+Be explicit and unambiguous in all responses:
+
+- Use literal language; avoid idioms, metaphors, or figurative expressions that could be misinterpreted
+- State assumptions explicitly rather than leaving them implicit
+- When suggesting multiple options, clearly label them and explain the trade-offs
+- If a request is ambiguous, ask specific clarifying questions before proceeding
+- Provide concrete examples when explaining abstract concepts
+- Break down complex tasks into explicit, numbered steps when helpful
```
pcanal, line 31: How does the agent know 'when' it is helpful?

.github/copilot-instructions.md
```diff
@@ -0,0 +1,100 @@
...
+- If you're uncertain about what the developer wants, state what you understand and ask for confirmation
+
+### External Resources
+
+When the developer provides HTTPS links in conversation:
+
+- You are permitted and encouraged to fetch content from HTTPS URLs using the `fetch_webpage` tool
+- This applies to documentation, GitHub issues, pull requests, specifications, RFCs, and other web-accessible resources
```
pcanal, line 39: What is the value added of this statement? I.e. Conversely this seems to imply that there is an unnamed set of URLs where even if the user is providing explicitly (as said on line 36) the URL, the agent is not allowed to go and fetch it.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,100 @@
...
+- If the link is not accessible or the content is unclear, report this explicitly
+- Always verify that fetched information is relevant to the developer's question before incorporating it into your response
+
+## Workspace Environment Setup
+
+### setup-env.sh Integration
+
+When working in this workspace, always source `setup-env.sh` before executing commands that depend on the build environment:
+
+- **Repository version**: `srcs/phlex/scripts/setup-env.sh` - Multi-mode environment setup for developers
```
pcanal, line 50: This script and README seems to yet be merge into the repository, is that correct?

.github/copilot-instructions.md
```diff
@@ -0,0 +1,100 @@
...
+- Without this flag, `find` will skip linked directories and miss significant portions of the codebase
+
+## Text Formatting Standards
+
+**CRITICAL: Apply to ALL files you create or edit (bash scripts, Python, C++, YAML, Markdown, etc.)**
+
+- All text files must end with exactly one newline character and no trailing blank lines
+- **Never add trailing whitespace on any line** (spaces or tabs at end of lines)
+- This includes blank lines - they should contain only the newline character, no spaces or tabs
+- Exception: Markdown two-space line breaks (avoid; use proper paragraph breaks instead)
```
pcanal, line 82: If there is a better alternative, why allow the lesser alternative?

pcanal (approved):
LGFM

.github/copilot-instructions.md
greenc-FNAL: I have asked Claude to streamline the instructions to minimize token use while retaining their efficacy. I'm not sure I'd try to gauge overlap or redundancy myself just based on the words used, since it's added reinforcement language in other places where what ws there was apparently not sufficient to achieve the desired effect.

.github/copilot-instructions.md
greenc-FNAL: Probability trees, token prediction, training, context. It doesn't "know" anything, it just predicts tokens. I know that seems glib, but this is all emergent behavior, and guessing things like this can lead to mistaken assumptions and over-reliance on anthropomorphization.

.github/copilot-instructions.md
greenc-FNAL: It often elaborates on a general point with specific examples, because it's not able to "reason" about whether something is in a particular class or not. It's not an exclusive list, just categorical clarification.

greenc-FNAL:
@pcanal As a general comment, I'd note that this entire document was authored by Claude, with the exception of a couple of conditionalizing tweaks from me. I would ask Claude to preserve something in its instructions in the best way for it to be able to follow them, and this is what resulted. Occasionally I would point out something it did that would seem to contradict the instructions, and ask it to adjust them so that it doesn't happen again. It's all about weights rather than absolutes, and as humans we can't reliably assess what weight any given word or phrase will have.

.github/copilot-instructions.md
pcanal: I am still not sure why a category is needed and whether this statement only allows (for better or worse) to fetch things that are within those categories.

.github/copilot-instructions.md
pcanal: It probably does not matter much (i.e. your preference wins), I would try to phrase it as follow though:
```suggestion
- Exception: Markdown use proper paragraph breaks unless the existing text already uses two-space line breaks.
```

.github/copilot-instructions.md
pcanal: I wonder if it makes an actual difference (`with` or `without` the `when helpful`).

PR #60, opened 2025-10-10
Update registration for transforms, folds, unfolds, and output modules

knoepfel:
This PR adjusts the interface for transforms, folds, and unfolds to be like (e.g.)

```c++
m.transform("MyTransform", f, concurrency::unlimited)
  .input_family("a"_in("spill"))
  .output_products("b");
```


test/demo-giantdata/unfold_transform_fold.cpp
```diff
@@ -152,14 +145,14 @@ int main(int argc, char* argv[])
          auto run_id = hwf.level_id().parent()->parent()->parent()->number();
          demo::accumulateSCW(scw, *hwf, run_id, subrun_id, spill_id, apa_id);
        },
-       concurrency::unlimited)
-      .fold("clamped_waves"_in("APA"))
-      .to("summed_waveforms")
-      .partitioned_by("spill") // partition the output by the spill
-      ;
+       concurrency::unlimited,
+       "spill" // partition the output by the spill
```
wddgit: Compared to CDR page 35. There is no initializer and the partition and concurrency arguments are reversed. I'm not sure whether this is a CDR or PR issue.

wddgit (approved):
The only thing I noticed other than what I already commented on was that page 25 and the little note on page 26 of the CDR need updates to be consistent. But that shouldn't affect the PR and I completed the review. Looks like an improvement.

test/demo-giantdata/unfold_transform_fold.cpp
knoepfel: This is due to a difficulty in what is allowed in C++ with specifying default arguments and variadic templates.  We should discuss this as a group.  I will create an issue regarding this that can be resolved after this PR goes in.

knoepfel: Issue noted at: https://github.com/Framework-R-D/phlex/issues/91

PR #65, opened 2025-10-15
Fix filtering race condition

knoepfel:
This PR fixes a data race in filtering, where a filter result was being sent multiple times for the same data cell.  This PR also introduces unit tests that verify the expected filtering behavior.

### DIFF
--- a/phlex/core/filter.cpp
+++ b/phlex/core/filter.cpp
@@ -4,7 +4,6 @@
 
 #include "fmt/std.h"
 #include "oneapi/tbb/flow_graph.h"
-#include "spdlog/spdlog.h"
 
 using namespace phlex::experimental;
 using namespace oneapi::tbb;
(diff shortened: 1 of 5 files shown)


greenc-FNAL:
This isn't the cause of the occasional test failure, is it?

knoepfel:
> This isn't the cause of the occasional test failure, is it?

Maybe.  I know at least one of the benchmark tests that was failing  (`benchmark:09`) was using filters.  I haven't seen it fail with this fix.  But I want to test it on `scisoftbuild02`; I've been testing on my Mac.

phlex/core/filter.cpp
```diff
@@ -81,8 +81,9 @@ namespace phlex::experimental {
       for (std::size_t i = 0ull; i != nargs_; ++i) {
         downstream_ports_[i]->try_put({stores[i], eom, msg_id});
       }
+      // Decision must be erased while access is claimed
```
wddgit: The bug fix itself seems good. Holding a lock while you erase the member is correct.

The other lock might be improved. I am not sure what exactly is being protected by the ```data_map::accessor a;``` but I think you might add a ```{``` before the line defining it and closing it ```}``` just before this line. Keeping that lock for less time might be a small improvement.

wddgit (approved):
Looks like it fixes the bug. There was one minor suggestion, but that is all I noticed.

PR #77, opened 2025-10-22
Update and use image; fix/improve clang-tidy use

greenc-FNAL:
- **Update CI image to include LLVM and use it in CI**
- **Ensure we detect clang-tidy failure**
- **AI-ish check specification**
- **Remove requirement for specific checks**
- **Provide a more informative summary on clang-tidy failure**
- **Fix configure command and avoid unnecessary `execute_process()`**
- **Remove unnecessary build**
- **Correct target name**
- **Update CI image**

### DIFF
--- a/ci/Dockerfile
+++ b/ci/Dockerfile
@@ -48,7 +48,7 @@ RUN <<EOF
   # Create and prepare phlex-development environment
 
   . spack/share/spack/setup-env.sh
-  spack versions -s llvm | grep -qEe '^[[:space:]]+21\.1\.3$' || spack checksum -ab llvm 21.1.3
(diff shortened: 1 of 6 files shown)


ci/Dockerfile
```diff
@@ -14,39 +14,97 @@ ARG parallelism=18
...
+
+RUN <<EOF
+  # Install Phlex' dependencies
+  . /entrypoint.sh
+  spack --timestamp install --fail-fast -j $parallelism -p 1 \
+    --no-add --only-concrete --only dependencies --no-check-signature \
+    phlex
+  # Clean up to reduce layer size
+  spack clean -dfs
+EOF
```
knoepfel: Why separate these two steps?  The first one appears to install everything that is not already installed; the second one will ensure that everything except `phlex` is installed.  The consequence is the same, no?

PR #92, opened 2025-10-31
Remove unused code when forming graph

knoepfel:
This code was filling a `std::map<...>` of consumed products, but the map is never used.  This PR removes that map, and it removes the lambda expression/closure object used to create the map.


beojan (approved):
Looks good. The removed map and lambda are indeed unused.

PR #93, opened 2025-10-31
Adjust header guards to use LLVM style

knoepfel:
The current `clang-tidy` configuration for Phlex requires macros to use `UPPER_CASE`, and the header guards are all `lower_case` right now.  If we switch to the LLVM header-guard style, that error goes away.

This PR does not make the corresponding changes to the `form` files.


aolivier23 (approved):
Changes all look consistent and limited to the scope of the PR.  Looks like a good change to me.

PR #94, opened 2025-11-04
Overhaul coverage tooling for GCC and Clang

greenc-FNAL:
- CreateCoverageTargets.cmake now drives gcov/lcov and llvm-cov with shared
  normalization, summary, and cleanup targets.
- Introduced PhlexTargetUtils helpers for target discovery reused by clang-tidy.
- Added coverage-gcc/coverage-clang presets, the coverage symlink helper script,
  a build-coverage default, and LLVM profile configuration for tests.
- Reworked scripts/coverage.sh to follow the new targets, support preset
  selection, copy artifacts, and generate LLVM summaries.
- Updated CI actions and workflows to choose compilers, respect the FORM toggle,
  gate artifact uploads, and package coverage bundles consistently.
- Extended ignore rules, added cetmodules to the Spack manifest, and corrected
  the FORM tech_setting forward declaration.

### DIFF
--- a/ci/spack.yaml
+++ b/ci/spack.yaml
@@ -2,7 +2,6 @@ spack:
   specs:
   - phlex
   - catch2
-  - cetmodules
   - cmake
   - lcov
   - ninja


knoepfel (commented):
Well done, @greenc-FNAL.  I am tempted to approve this as is in the interest of moving forward.  I do have a couple thoughts, though, for future adjustment to CI:

1. If we know that GCC code coverage is unreliable and Clang is sufficient for our code-coverage needs, I'd remove the GCC code-coverage-related code.  We can make this a separate GitHub issue.
2. One of our goals is to be able to build Phlex for both GCC and Clang.  You've now demonstrated that is possible.  Do we now have separate GitHub checks for the GCC build _and_ the Clang build (which also handles the code coverage)?

form/persistence/persistence.hpp
```diff
@@ -16,7 +16,7 @@
 // forward declaration for form config
 namespace form::experimental::config {
   class output_item_config;
-  class tech_setting_config;
+  struct tech_setting_config;
```
knoepfel: Is this adjustment necessary for this PR?

form/persistence/ipersistence.hpp
```diff
@@ -9,7 +9,7 @@
 
 namespace form::experimental::config {
   class output_item_config;
-  class tech_setting_config;
+  struct tech_setting_config;
```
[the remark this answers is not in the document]
greenc-FNAL: It's necessary for compilation with `clang++`, so yes. I can put it through as a separate PR if you like, but it is a prerequisite for successful coverage testing with LLVM/Clang.

greenc-FNAL:
> 1\. If we know that GCC code coverage is unreliable and Clang is sufficient for our code-coverage needs, I'd remove the GCC code-coverage-related code.  We can make this a separate GitHub issue.

GCC code coverage is unreliable _at the moment_, certainly, although `gcovr` does omit the problematic branch data from the Codecov upload.

I would prefer to remove the GCC/`gcov` functionality as a separate PR, so that reverting it would be a reasonable starting point for restoring the functionality at a later date.

form/persistence/persistence.hpp
greenc-FNAL: See https://github.com/Framework-R-D/phlex/pull/94#discussion_r2500131167

greenc-FNAL:
> 2\. One of our goals is to be able to build Phlex for both GCC and Clang.  You've now demonstrated that is possible.  Do we now have separate GitHub checks for the GCC build _and_ the Clang build (which also handles the code coverage)?

We don't have an automatic CI step for checking the GCC build after this PR is merged, although adding one would be trivial. Would you like me to add it to this PR, or submit a separate one after this is merged?

knoepfel:
> Also: it is possible, using the Codecov concept of "[flags](https://docs.codecov.com/docs/flags)", to separate coverage reporting for FORM and Phlex, with different targets, etc. I'd be happy to look into that for a future PR.

That's reasonable.  It's also possible to get a view of the code coverage for individual subdirectories (`phlex/phlex` vs. `phlex/form`) without using flags.  But flags might be a better solution.

PR #100, opened 2025-11-07
Re-work image generation

greenc-FNAL:
Use a multi-target `Dockerfile` instead of `ci/Dockerfile` being the basis of a `FROM ...` for a dev image.

This avoids ownership/uid/permission/Spack cache issues in both CI and dev containers, while maximizing both Dockerfile maintainability/reuse, and image layer reuse.

### DIFF
--- a/ci/Dockerfile
+++ b/ci/Dockerfile
@@ -72,7 +72,6 @@ COPY entrypoint.sh /entrypoint.sh
 RUN <<'CONFIGURE_SPACK_DEFAULTS'
 set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


greenc-FNAL:
Use a multi-target `Dockerfile` instead of `ci/Dockerfile` being the basis of a `FROM ...` for a dev image.

This avoids ownership/uid/permission/Spack cache issues in both CI and dev containers, while maximizing both Dockerfile maintainability/reuse, and image layer reuse.

PR #119, opened 2025-11-17
Add type checking to graph construction

beojan:
Add type checking to products (at least when registering algorithms in C++).

The `qualified_name` for output products, and the `specified_label` for input products now have type information attached (in the form of a custom `type_id` currently calculated mostly at compile time) that is compared in `find_products`. 

I had to use a custom `type_id` so we could do things like treating `handle<int>`, `int const&`, and `int const*` as the same type. This should also simplify extending this type checking to Python algorithms.

I think I've done the best I can with code coverage. I don't know why it thinks the formatting code isn't covered, it definitely should be in my `type_id.cpp` test.

### DIFF
--- a/phlex/core/edge_creation_policy.cpp
+++ b/phlex/core/edge_creation_policy.cpp
@@ -31,9 +31,7 @@ namespace phlex::experimental {
             spdlog::debug("Matched {} ({}) from {} and types match. Keeping in candidate list.",
                           spec.full(),
                           query.to_string(),
-                          producer.node.full(),
-                          spec.type(),
-                          producer.type);
+                          producer.node.full());
           } else {
             spdlog::warn("Matched {} ({}) from {} and types match, but not exactly (produce {} and "
                          "consume {}). Keeping in candidate list!",


phlex/model/type_id.hpp
knoepfel: I agree with the spirit of what this file is doing:

```
// This is a type_id class to store the "product concept"
// Using our own class means we can treat, for example, all "List"s the same
```

I should say, however, that it's not obvious to me whether the framework should be responsible for calculating what the product concept should be, given a C++ type.  I would have expected that using (e.g.) `std::type_index` would have been largely sufficient for the first prototype, although it would be clearly inadequate for what we eventually want to do.

Do you think this approach is readily extensible in supporting Python product types and their corresponding product concepts?

phlex/model/type_id.hpp
```diff
@@ -0,0 +1,308 @@
...
+    enum class builtin : unsigned char {
+      void_v = 0, // For completeness
+      bool_v = 1,
+      char_v = 2,
+      int_v = 3,
+      short_v = 4,
+      long_v = 5,
+      long_long_v = 6,
+      float_v = 7,
+      double_v = 8
```
knoepfel: Looks like we're missing `long double`

knoepfel (commented):
@beojan, bravo.  I think you have hit all of the spots in the implementation that would be affected by imposing type constraints when establishing edges.  I have several comments below for your consideration and for further discussion.  A great start—thanks.

phlex/model/type_id.hpp
beojan: I think it should be pretty easily extensible to Python product types. We'd need a Python version of the code to create the `type_id` that reflects whatever correspondence we decide on between C++ and Python types.

beojan:
I ran the coverage script locally using the LLVM backend and that shows line coverage for `type_id.hpp` at 96.9%

image

knoepfel (approved):

PR #129, opened 2025-11-21
Reorganize and refactor workflows and actions

greenc-FNAL:
- **feat(workflows): Reorganize and standardize GitHub Actions**
- **refactor(workflows): align cmake-build with other workflows**

### DIFF
--- a/.github/workflows/clang-format-check.yaml
+++ b/.github/workflows/clang-format-check.yaml
@@ -7,31 +7,34 @@ permissions:
 
 on:
   pull_request:
-    branches: [ main ]
+    branches: [ main, develop ]
   workflow_dispatch:
 
 jobs:
-  detect-clang-format-changes:
+  pre-check:
+    runs-on: ubuntu-latest
+    outputs:
+      is_act: ${{ steps.detect_act.outputs.is_act }}
+    steps:
+      - name: Detect act environment
+        id: detect_act
+        uses: Framework-R-D/phlex/.github/actions/detect-act-env@main
+
+  detect-changes:
+    needs: pre-check
+    if: github.event_name != 'workflow_dispatch' && needs.pre-check.outputs.is_act != 'true'
     runs-on: ubuntu-latest
     permissions:
       contents: read
       packages: read
-
     outputs:
       has_changes: ${{ steps.filter.outputs.matched }}
-      changed_files: ${{ steps.filter.outputs.matched_files }}
-      is_act: ${{ steps.detect_act.outputs.is_act }}
-
     steps:
     - name: Checkout code
       uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
       with:
         fetch-depth: 0
         path: phlex-src
-
-    - name: Detect act environment
-      id: detect_act
-      uses: Framework-R-D/phlex/.github/actions/detect-act-env@main
 
     - name: Detect C++ formatting changes
       id: filter
@@ -53,8 +56,11 @@ jobs:
         fi
 
   clang-format-check:
-    needs: detect-clang-format-changes
-    if: ${{ (needs.detect-clang-format-changes.outputs.has_changes == 'true') || (github.event_name == 'workflow_dispatch') || (needs.detect-clang-format-changes.outputs.is_act == 'true') }}
+    needs: [pre-check, detect-changes]
+    if: >
+      github.event_name == 'workflow_dispatch' ||
+      needs.pre-check.outputs.is_act == 'true' ||
+      (needs.detect-changes.result == 'success' && needs.detect-changes.outputs.has_changes == 'true')
     runs-on: ubuntu-latest
     permissions:
       contents: read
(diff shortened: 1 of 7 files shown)


greenc-FNAL:
Note: This has been verified on [my fork](https://github.com/greenc-FNAL/phlex/tree/main), but [2adf3a3](https://github.com/Framework-R-D/phlex/pull/129/commits/2adf3a3c44a2265c33eb0a935ccde2e1cc847433) causes tests to fail here because it refers safely to reusable actions that are not on [this repository's `main`](https://github.com/Framework-R-D/phlex/tree/main) yet.

marcpaterno (commented):
Here are comments from GitHub Copilot that seem work considering:

Things I found questionable, fragile or worth double-checking

Word-splitting / globbing risk when appending EXTRA_OPTIONS to CMAKE_ARGS:

The configure-cmake action does: CMAKE_ARGS+=($EXTRA_OPTIONS). If EXTRA_OPTIONS contains quoted values, spaces, or shell metacharacters, this will lead to word-splitting and globbing. Prefer to parse safely (e.g. read into an array with proper splitting rules, or pass EXTRA_OPTIONS as a single string and use eval-safe handling).
Mixed behavior for clang-tidy run: build step forced to succeed

The clang-tidy-run step was changed to wrap the cmake build & then append a trailing "true", which guarantees that step always returns success. This hides the build result from Actions and defers failure detection to later steps. That is acceptable if intentional, but it changes failure semantics and can hide pipeline status in logs (the job will still fail later if checks detect issues). Consider adding clear log lines and an explicit check step that fails with a descriptive error to make root cause easier to find.
Replacing env.ACT checks with outputs.is_act requires detect-act step be present and run before referencing it

The new pattern uses needs.detect-xxx.outputs.is_act; make sure the detect-act action is always executed in jobs that rely on its output. I saw detect-act additions in many jobs, but double-check every job that now references .outputs.is_act to ensure the detect-act step actually runs in that job (no cross-job output confusion).
Potential missing quoting in YAML expression references

Some expressions like base-ref: ${{ github.event_name == 'issue_comment' && fromJSON(steps.pr.outputs.result).base_sha || github.event.pull_request.base.sha || github.event.before }} are compact and rely on JSON parsing and conditional evaluation inside YAML. These can be fragile; test comment-triggered runs to ensure the expressions resolve correctly for all event types.
get-pr-info action uses actions/github-script and core.setOutput in a script block

The action uses actions/github-script to call github.rest.pulls.get and then core.setOutput in a script literal — ensure the version used supports this script shape. The file name is action.yml and this action is used via ./actions/get-pr-info which should work, but be cautious with version compatibility.
Using continue-on-error on lint steps but then failing in an evaluate step

This is a fine pattern (collects output then fails in a dedicated evaluation step), but if you ever remove the evaluator or its if: always() condition, the lint failure would become a soft failure. Keep the evaluator in place and make it clear in logs why failure is declared.
In several places actions/checkout and other actions are pinned to commit ids rather than tags; the commit ids differ across files. Pins are OK, but keep a plan for updating all pins consistently (some places still used old pins before this patch).

Removal of some if: always() wrappers

I noticed a few if: always() conditions were removed (e.g. Unknown Compiler Error step). Removing always() can alter whether diagnostic steps run on success vs failure; double-check intended behavior so diagnostics still run when you expect them to.

greenc-FNAL:
Based on the AI comments @marcpaterno [posted above](https://github.com/Framework-R-D/phlex/pull/129#pullrequestreview-3501316981), I have checked over the workflows with respect to the issues flagged, and I have made some changes with 21fd9b7 to further improve the structure, clarity, efficiency and consistency of the workflows. From the commit message:

> refactor(workflows): Standardize and improve CI workflows
>
> This commit refactors the GitHub Actions workflows to improve consistency, efficiency, and clarity. The changes ensure that all workflows follow a standard structure and provide better feedback to developers.
>
> Key improvements include:
>
> - **Structural Consistency:**
>   - A `pre-check` / `detect-changes` job pattern has been implemented across all relevant workflows to standardize the workflow structure.
>   - Job names have been renamed for consistency (e.g., `tidy-pre-check` is now `pre-check`).
>
> - **Efficiency:**
>   - Workflows now skip the `detect-changes` step when triggered by `workflow_dispatch` or run locally with `act`, avoiding unnecessary work.
>   - The `if` conditions on all downstream jobs have been updated to correctly handle cases where the detection job is skipped or has no changes.
>
> - **Improved Feedback and Logging:**
>   - The output messages in `*-skipped` jobs and other informational steps are updated to use the `::notice::` and `::error::` annotations for better visibility and consistency.
>   - The bot command for `clang-tidy-fix` has been changed to `@phlexbot tidy-fix` to align with the `<subject>-<verb>` convention.
>
> - **CodeQL Analysis Enhancements:**
>   - The `codeql-analysis` workflow has been updated to provide clear, actionable feedback for pull requests from forks.
>   - If new alerts are found, the workflow now logs the full details using `::error::` and fails the run.
>   - If only fixed alerts are found, the details are logged with `::notice::` and the run succeeds.

PR #141, opened 2025-12-02
Some FORM fixes

knoepfel:
This PR:

- Links the `ROOT::RIO` library during the dictionary-generation step
- Uses `snprintf` instead of `sprintf` when filling a string buffer (per apple-clang warning)

### DIFF
--- a/test/form/data_products/CMakeLists.txt
+++ b/test/form/data_products/CMakeLists.txt
@@ -3,7 +3,7 @@ add_library(
   ${FORM_DATA_PROD_LIB_NAME} SHARED track_start.cpp
   ) # REFLEX_GENERATE_DICTIONARY doesn't work trivially without making a shared
     # library
-target_link_libraries(${FORM_DATA_PROD_LIB_NAME})
+target_link_libraries(${FORM_DATA_PROD_LIB_NAME} ROOT::RIO)
 
 if(FORM_USE_ROOT_STORAGE)
   find_package(ROOT REQUIRED COMPONENTS Core RIO)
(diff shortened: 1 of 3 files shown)


test/form/data_products/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ add_library(
   ${FORM_DATA_PROD_LIB_NAME} SHARED track_start.cpp
   ) # REFLEX_GENERATE_DICTIONARY doesn't work trivially without making a shared
     # library
-target_link_libraries(${FORM_DATA_PROD_LIB_NAME})
+target_link_libraries(${FORM_DATA_PROD_LIB_NAME} ROOT::RIO)
```
gemmeren: I was wondering whether the `ROOT::RIO` addition could be hidden behind a `if(FORM_USE_ROOT_STORAGE)`

knoepfel: Sorry, Peter.  Just merged this PR.  I made the assumption that `form_test_data_products` corresponded to the dictionary library.  I think I can do what you suggest...I'll create another PR.

aolivier23: I think it's hidden in the next level up here: https://github.com/Framework-R-D/phlex/blob/6838973ef28a22a3193ed7b15248f1eb2b48c0e9/test/form/CMakeLists.txt#L3

aolivier23: Even if I'm right, we might still want finer-grained control of when FORM uses ROOT.  Right now, I think our tests rely on having a ROOT back end.

Issue #90, opened 2025-10-30
Unfold interface should reflect design document


knoepfel:
[As observed](https://github.com/Framework-R-D/phlex/pull/60#discussion_r2479301096) by @wddgit, the interface to `unfold` should require a name as the first argument.

PR #161, opened 2025-12-04
Make unfold interface reflect design document

wddgit:
Make unfold interface reflect design document. Name should be the first argument of the unfold function.

See Issue #90


wddgit:
Note that the code coverage dropped because I deleted code that I presume had coverage. I didn't add any new code to glue.hpp.

knoepfel:
> Note that the code coverage dropped because I deleted code that I presume had coverage. I didn't add any new code to glue.hpp.

Correct.

PR #166, opened 2025-12-05
Add `product_store::size()` accessor

knoepfel:
This is a necessary addition to allow FORM to reserve a vector size that equals the number of products in the store (e.g.):

```cpp
void save_to_output(product_store const& store)
{
  std::vector<some_type> products_to_save;
  products_to_save.reserve(store.size());
  // ...
}

### DIFF
--- a/phlex/model/products.hpp
+++ b/phlex/model/products.hpp
@@ -98,6 +98,7 @@ namespace phlex::experimental {
     const_iterator begin() const noexcept;
     const_iterator end() const noexcept;
     size_type size() const noexcept;
+    bool empty() const noexcept;
 
   private:
     static void throw_mismatched_type [[noreturn]] (std::string const& product_name,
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: phlex/model/products.hpp [hunk 1]
```diff
@@ -96,6 +97,7 @@ namespace phlex::experimental {
     bool contains(std::string const& product_name) const;
     const_iterator begin() const noexcept;
     const_iterator end() const noexcept;
+    size_type size() const noexcept;
```
pcanal:
```suggestion
    size_type size() const noexcept;
    void empty() const noexcept { return products_.empty(); }
```

pcanal (dismissed):
LGTM.

phlex/model/products.hpp
knoepfel: Implemented (but with a `bool` return type).

PR #168, opened 2025-12-05
Convert FORM tests to use Cetmodules

greenc-FNAL:
- Convert test/form tests to use cet_test().

- Update writer.cpp and reader.cpp to accept filename argument.

- Pass shared data file path to tests to support parallel execution.

Run `cmake-format` (FORM)


aolivier23 (approved):
Looks reasonable to me.   Thank you for doing this!  I'm trusting that `cet_test()` does roughly what the old CMake code did as I'm not familiar with how the cet macros work.

gemmeren (approved):
Thanks Chris, looks good.

PR #177, opened 2025-12-09
Add FORM output module for phlex product_store integration

barnaliy:


form/CMakeLists.txt
```diff
@@ -18,11 +18,21 @@ endif()
...
 endif()
+
+add_library(form_module form_module.cpp)
+
+target_link_libraries(form_module PRIVATE 
+    phlex::module
+#    phlex::experimental
+)
+
+target_link_libraries(form_module PRIVATE phlex::module phlex::experimental)
```
gemmeren: I believe you need `phlex::module` and `phlex::model` not `phlex::experimental`, which I don't believe exists. (also, lines 31... and 36 are ~duplicate)

barnaliy: Thanks! The duplicate lines have been removed.

form/form_module.cpp
```diff
@@ -0,0 +1,141 @@
...
+    {
+      // Check if store is empty - smart way, check store not products vector
+      if (store.empty()) {
+        return;
+      }
+
+      // STEP 1: Extract metadata from Phlex's product_store
+
+      // Extract creator (algorithm name)
+      std::string creator = store.source();
```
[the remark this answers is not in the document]
pcanal: That is a good start (see test/product_store.cpp).  We should also have an example that put it in play in the regular workflow  i.e. register it and then trigger the workflow, but maybe we should leave this 2nd part for a follow up PR.

form/CMakeLists.txt
gemmeren: Thanks! Works now.

gemmeren (approved):

form/CMakeLists.txt
```diff
@@ -18,11 +18,16 @@ endif()
...
-add_subdirectory(mock_phlex)
 add_subdirectory(core)
 add_subdirectory(util)
 add_subdirectory(persistence)
 add_subdirectory(storage)
 if(FORM_USE_ROOT_STORAGE)
   add_subdirectory(root_storage)
 endif()
+
+add_library(form_module form_module.cpp)
```
knoepfel:
```suggestion
add_library(form_module MODULE form_module.cpp)
```

knoepfel:
@barnaliy, the test now found the file, but there's another error:

```console
 49/52 Test #52: job:form_module .........................***Failed    0.14 sec
  Using configuration file: /__w/phlex/phlex/phlex-src/test/form/form_test.jsonnet
  [2025-12-12 22:20:18.332] [info] Number of worker threads: 4
  [2025-12-12 22:20:18.332] [debug] Created provider node add:provide_i making output i ϵ event
  [2025-12-12 22:20:18.332] [debug] Created provider node add:provide_j making output j ϵ event
  /__w/phlex/phlex/phlex-build/bin/phlex: symbol lookup error: /__w/phlex/phlex/phlex-build/libform_module.so: undefined symbol: _ZN4form12experimental14form_interfaceC1ESt10shared_ptrINS0_18product_type_namesEERKNS0_6config18output_item_configERKNS5_19tech_setting_configE
```

This means that the `form_module` library is missing some link-time libraries.  You need to add the correct FORM libraries that the code depends upon to this line:

```cmake
target_link_libraries(form_module PRIVATE phlex::module 
 # put required FORM library/libraries here ...
)
```

form/form_module.cpp
```diff
@@ -0,0 +1,145 @@
...
+    {
+      std::cout << "FormOutputModule initialized\n";
+      std::cout << "  Output file: " << m_output_file << "\n";
+      std::cout << "  Technology: " << m_technology << "\n";
+
+      // Build FORM configuration
+      form::experimental::config::output_item_config output_cfg;
+      form::experimental::config::tech_setting_config tech_cfg;
+
+      // Register products that will be written
```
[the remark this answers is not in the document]
gemmeren: Thanks Kyle,

should we create a git-issue documenting this, or add a FIXME in the code (I know people hate those, you are supposed to;)? 

IIRC, FORM had gotten products to write from a mock config, but it should be rather straight-forward to enable it to write all products and algorithm has.

Peter

gemmeren:
Hi Barnali,
the problem is due to not having a dictionary. you will need to create one for sum, just like we do with trackStart.

gemmeren:
Hi Barnali, Philippe,

my concerns about the dictionary came from a prior test fail: [cmake fail](https://github.com/Framework-R-D/phlex/actions/runs/20277088117/job/58228789728?pr=177) 

```
  Error: 2-16 17:41:38.773] [error] ROOT_TBranch_ContainerImp::setupWrite unsupported type: 
  terminate called after throwing an instance of 'std::runtime_error'
    what():  ROOT_TBranch_ContainerImp::setupWrite unsupported type: 
  /__w/phlex/phlex/phlex-build/_deps/cetmodules-src/libexec/cet_exec_test: line 133:  2677 Aborted                 (core dumped) "$@"
```

so here, it throws in [tbranch write](https://github.com/Framework-R-D/phlex/blob/243e535838369724956eaa624c021d20815ad6be/form/root_storage/root_tbranch_container.cpp#L75)

was that a red herring or has it been addressed?

Peter

barnaliy:
@pcanal :  
The Mystery:
PC's local: Plugins load YES, but ROOT file ERROR
GitHub Actions: Plugins DON'T load 

Same code, different environments, different errors!

Does this suggest 
The PHLEX_PLUGIN_PATH issue might be environment-specific to GitHub Actions. The path that works for Philippe locally doesn't work in the CI environment ?

What PHLEX_PLUGIN_PATH are you using? The CI still can't find the "source" plugin even with CMAKE_BINARY_DIR:CMAKE_BINARY_DIR/form.

pcanal:
I pushed a commit with the solution to the plugin issue that was provided by Kyle.

there is no source plugin anymore.  The source configuration in [form_test.jsonnet](https://github.com/Framework-R-D/phlex/blob/7fd6f3ee5cd92e1b744bb063ebff4f69093b3347/test/form/form_test.jsonnet):
```
  source: {
    plugin: 'source',
    max_numbers: 10,
  },
```
should be replaced with:
```
  source: {
    plugin: 'generate_layers',
    layers: {
      event: { total: 10 }
    }
  },
  ```

**Comment by pcanal** [comment]
> that should be done here [setDirectory](https://github.com/Framework-R-D/phlex/blob/3aa9fdab820d51dce4f52ee9636cb813cc6d99e6/form/root_storage/root_ttree_container.cpp#L45)

Indeed.

> I suspect that maybe the file wasn't setup (e.g. failing to pass its name?)

It is setup and correct at the time of the SetDirectory, so something else is going awry.

**Comment by pcanal** [comment]
With this the situation is:

1. - the build succeed and pass the tests.
2. - the test is actually known to not test whether the data is stored properly.
3. The code coverage is essentially as expected.  Only 3 lines are not covered:
- - 2 error cases path
- - RNTuple support.

I think (assuming the CI pass as expected) that we can merge but the next step is to write an additional test to verify that the content of output.root is what it is meant to be.

**Comment by gemmeren** [comment]
Thanks @pcanal ,

for all your help. I agree we should add a simple file content check, that does not need to include Phlex/FORM, but could be a simple file open, tree->Print(), comparison.

**Comment by gemmeren** [comment]
All good to move forward, just minor comments/questions about reading.

**Comment by gemmeren** [comment]
@pcanal  `scan()` is fine by me, `Print()` is good too, since it shows sizes, compression, which are unlikely to match the reference if the data differs or disappears...

**Review by pcanal — CHANGES_REQUESTED** [review]
https://github.com/Framework-R-D/phlex/pull/177/changes#r2628320178
https://github.com/Framework-R-D/phlex/pull/177/changes#r2628322978

must be addressed before merging.

**Comment by barnaliy** [comment]
@pcanal : 

> https://github.com/Framework-R-D/phlex/pull/177/changes#r2628320178 https://github.com/Framework-R-D/phlex/pull/177/changes#r2628322978
> 
> must be addressed before merging.

The link takes me to the top of the PR. Could you please point out the right file and line number?

**Comment by pcanal** [comment]
@barnaliy @gemmeren I added another commit https://github.com/Framework-R-D/phlex/pull/177/changes/9b49214c410d574b384e546b34631bf4245b05f8 to extent the list of support types.

If we can address https://github.com/Framework-R-D/phlex/pull/177#discussion_r2628823500 then we have enough in place to support arbitrary simple configured products (instead of hard coding a single product).

**Comment by gemmeren** [comment]
Thanks Philippe,
for adding the basic/native types. @barnaliy  Maybe bool, char and string can be added as well (can be separate PR).

**Comment by knoepfel** [comment]
@phlexbot format

File: form/form/form.hpp [hunk 5]
```diff
@@ -4,28 +4,56 @@
...
+  // Type map - moved from mock_phlex (FORM needs this)
+  struct product_type_names {
+    std::unordered_map<std::type_index, std::string> names;
+  };
+
+  // Product structure - moved from mock_phlex, but without id field
+  struct product_with_name {
+    std::string label;    // product name
+    void const* data;     // pointer to actual data
+    std::type_index type; // type information
```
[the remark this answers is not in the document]
knoepfel: Issue https://github.com/Framework-R-D/phlex/issues/199 (created by @pcanal) covers this.

knoepfel:
@barnaliy and @pcanal, we now have successful builds and tests.  Does anything else need to be done before approving and merging this PR?

Issue #108, opened 2025-11-12
Adjust product specification to allow creator names


knoepfel:
The goal is to permit registration code like:

```.cpp
g.transform("A", make_an_a).input_family(...);
g.transform("B", make_a_b).input_family("A"_in("spill"));
```

where the output name of the product created by transform "A" does not need to be explicitly specified for transform "B".  This means that the specified name corresponds to the node name and not the name of the data product.  The name of the data product only needs to be specified to disambiguate between other products produced by the same creator _with_ the same type.

PR #184, opened 2025-12-11
FEAT: Specify input product using producer name

beojan:
This is a preliminary solution for #108. Output products still need to be specified as empty strings at this point.

The API for input product specification without a product name is:
```c++
.input_products(from(producer_name).in(layer_name), ...)
```


beojan:
Closing this PR. Since the prototype has been released we've come up with a more radical change with a new and improved API.

PR #186, opened 2025-12-12
Enable direct conversion from configuration parameter to product_query

### LINKED ISSUE #163: Enable direct conversion from configuration parameter to `product_query`
Users should be able to specify a configuration like

```jsonnet
module_name: {
  plugin: "<plugin path>",
  input_product: { product: "a", in: "spill"}
}
```

and then, in the registration code, be able to directly express the value of the `input_product` parameter above as a `product_query` object:

```cpp
PHLEX_REGISTER_ALGORITHMS(m, config)
{
  m.transform(...).input_family(config.get<product_query>("input_product")); 

  // which is equivalent to:
  //
  // m.transform(...).input_family(
  //   product_query{config.get<std::string>("input_product.product"),
  //                 config.get<std::string>("input_product.in")}
  // );
}
```

knoepfel:
Closes #163.

### DIFF
--- a/phlex/configuration.hpp
+++ b/phlex/configuration.hpp
@@ -9,6 +9,17 @@
 #include <vector>
 
 namespace phlex::experimental {
+
+  namespace detail {
+    template <typename T>
+    T value_decorate_exception(boost::json::object const& obj, std::string const& key)
+    try {
+      return value_to<T>(obj.at(key));
+    } catch (std::exception const& e) {
+      throw std::runtime_error("Error retrieving parameter '" + key + "':\n" + e.what());
+    }
+  }
+
   class configuration {
   public:
     configuration() = default;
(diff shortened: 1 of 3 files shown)


knoepfel:
@phlexbot format

phlex/configuration.hpp
```diff
@@ -32,10 +36,29 @@ namespace phlex::experimental {
       return get_if_present<T>(key).value_or(std::forward<T>(default_value));
     }
 
+    std::vector<std::string> keys() const
+    {
+      std::vector<std::string> result;
+      result.reserve(config_.size());
+      for (auto const& element : config_) {
+        result.push_back(element.key());
```
beojan: Could use `std::ranges::copy` and `std::ranges::views::keys`.

PR #187, opened 2025-12-12
expose the kind of json values to simplify the retrieval Python-side

wlav:
Adds a basic internal kind retrieval function, which greatly simplifies lookup and type conversion in the Python configuration wrapper. (Note that this method is meant to be temporary and only for the prototype. It will change once the configuration interface itself gets fleshed out.)

Edit: I saw the test coverage fail, but it's tested by the Python code which is waiting for this to be included first.

### DIFF
--- a/test/python/driver.cpp
+++ b/test/python/driver.cpp
@@ -6,9 +6,10 @@
 namespace {
   class number_generator {
   public:
-    number_generator(phlex::experimental::configuration const& config) :
-      n_{config.get<int>("max_numbers")}
+    number_generator(phlex::experimental::configuration const& config) n_{0}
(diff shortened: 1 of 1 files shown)


knoepfel:
@wlav, do you intend to add a unit test to cover the lines that are missing code coverage?

wlav:
I can look if there's an easy way of adding a C++ test or combine this change with the larger Python PR.

wlav:
I tried to temporarily add a use to one if the python tests, but it didn't like that as that test is apparently not run in the CI. I'm removing this PR and will make the change part of the larger Python PR, where it is tested.

PR #196, opened 2025-12-16
Specify plugin language as `cpp` or `py`

### LINKED ISSUE #164: Signify plugin language via `cpp` or `py` parameter
Currently, Phlex plugins are loaded via the syntax:

```jsonnet
modules: {
  a: {
    plugin: "<plugin path>"
  },
  b: {
    plugin: "<plugin path>"
  },
  ⋮
}
```

with the assumption that the loaded module is a C++ module.  To make loading C++ and Python modules symmetric, the recommended schema is instead:

```jsonnet
modules: {
  a: {
    cpp: "<shared object-file path>"
  },
  b: {
    py: "<python module>"
  },
  ⋮
}
```

Eventually, this should be replaced with a system that can calculate whether the plugin is C++ or Python.

knoepfel:
Enacts the following change for C++ plugins:

```diff
<plugin_label> : {
-  plugin: "my_cpp_plugin"
+  cpp: "my_cpp_plugin"
}
```

and the following change for Python plugins:

```diff
<plugin_label> : {
-  plugin: "pymodule"
-  pyplugin: "my_python_module.py"
+  py: "my_python_module"
}
```

Note that, for Python plugins, the `cpp: "pymodule"` configuration is automatically inserted into the configuration if the `py: ...` parameter is present.

---

- Closes #164.

### DIFF
--- a/test/adjust_config.cpp
+++ b/test/adjust_config.cpp
@@ -16,22 +16,22 @@ TEST_CASE("Adjust empty config", "[config]")
 TEST_CASE("Adjust config with py only", "[config]")
 {
   boost::json::object obj;
-  obj["py"] = "my_python_module.py";
+  obj["py"] = "my_python_module";
 
   auto const config = adjust_config("", std::move(obj));
-  CHECK(config.at("py").as_string() == "my_python_module.py");
+  CHECK(config.at("py").as_string() == "my_python_module");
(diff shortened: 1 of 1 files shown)


knoepfel:
@phlexbot format

marcpaterno (approved):

PR #205, opened 2025-12-17
Safer `product_query` construction

knoepfel:
Either default-construct `product_query` or require the constructor with both the `product_specification` and the `layer` argument.

Also switch to using default comparisons

### DIFF
--- a/phlex/core/product_query.hpp
+++ b/phlex/core/product_query.hpp
@@ -10,6 +10,8 @@
 namespace phlex {
   class product_query {
   public:
+    // FIXME: Boost JSON's parameter retrieval facilities require a default constructor
+    //        whenever the type is (e.g.) std::array<product_query, N>.
     product_query();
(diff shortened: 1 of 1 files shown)


phlex/core/product_query.hpp
```diff
@@ -3,22 +3,30 @@
...
+  class product_query {
+  public:
+    // FIXME: Boost JSON's parameter retrieval facilities require a default constructor
+    //        whenever the type is (e.g.) std::array<product_query, N>.
+    product_query();
+    product_query(experimental::product_specification spec, std::string layer);
+
+    auto const& spec() const noexcept { return spec_; }
+    auto const& layer() const noexcept { return layer_; }
+    void set_type(experimental::type_id&& type) { spec_.set_type(std::move(type)); }
```
beojan: Don't need the `std::move` here, it's already an rvalue reference.

beojan (dismissed):
Can't see any issues.

PR #225, opened 2026-01-07
Convert Boost JSON string to `std::string_view`

knoepfel:
The `fmt::format(...)` function can represent a Boost JSON string as an array of `char`s instead of character string (e.g.):"

```
Both 'cpp' and 'py' parameters specified for malformed3
    - cpp: ['m', 'y', '_', 'o', 't', 'h', 'e', 'r', '_', 'p', 'y', 't', 'h', 'o', 'n', '_', 'p', 'h', 'l', 'e', 'x', '_', 'm', 'o', 'd', 'u', 'l', 'e']
```

This PR makes use of Boost JSON string's implicit conversion operator to `std::string_view` so that `fmt::format(...)` will form the output correctly (e.g.):

```
Both 'cpp' and 'py' parameters specified for malformed3
    - cpp: my_other_python_phlex_module
```

### DIFF
--- a/phlex/app/load_module.cpp
+++ b/phlex/app/load_module.cpp
@@ -11,6 +11,7 @@
 
 #include <functional>
 #include <string>
+#include <string_view>
 
 using namespace std::string_literals;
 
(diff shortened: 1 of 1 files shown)


pcanal (approved):
LGTM (also verified it solves the issue in my configuration).

PR #247, opened 2026-01-09
Increase ruff's target python version to 3.12

wlav:
Increase ruff's target python version to 3.12. The default is 3.10, but out minimum required Python version (in spack etc.) is 3.12. This simplifies code development by allowing 3.12 language constructs.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,7 @@ warn_unreachable = true
 disable_error_code = "annotation-unchecked"
 
 [tool.ruff]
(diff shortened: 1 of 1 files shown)


wlav:
> FYI: it's more straightforward to add `target-version` to `pyproject.toml`

Found it, and it says 3.10 indeed. :) But it also affects the python files in the test directory which aren't part of any project.

wlav:
I started over. :) Let's try this one simple change and see how that goes.

Issue #199, opened 2025-12-17
`product_base` needs to record `std::type_info`


pcanal:
`product_base` currently records and returns only the `std::type_index` of the underlying data.  However one can not recover the `std::type_info` from the `std::type_index` (only the reverse is currently supported by C++).

In order for the I/O to know which type is being used, passing the `std::type_info` is much more convenient (one can demangle the name of the type info to get the C++ spelling of the type and `ROOT` records/remember the `type_info`).

Alternatively the I/O would have to be passed a `map<std::type_index, ....>` to understand the type_index (and/or ROOT dictionary would have to be enhanced to also record the type_index).

pcanal:
This was resolved by #252

PR #252, opened 2026-01-13
Update how type information is passed from Phlex to FORM

pcanal:
Given that ROOT knows about the type via their `std::type_info` (and their C++ names) and that we can not lookup a `std::type_info` from an `std::type_index` (but the reverse is supported by the C++ standard), we need to pass around the `type_info` (otherwise we need to build a map of supported types).

This solves https://github.com/Framework-R-D/phlex/issues/199

### DIFF
--- a/form/form_module.cpp
+++ b/form/form_module.cpp
@@ -38,8 +38,7 @@ namespace {
       }
 
       // Initialize FORM interface
-      m_form_interface =
-        std::make_unique<form::experimental::form_interface>(output_cfg, tech_cfg);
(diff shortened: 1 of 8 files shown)


knoepfel (dismissed):
Changes look good to me, @pcanal.  I defer to @gemmeren and @barnaliy for their comments on the FORM changes.

barnaliy:
@pcanal : This is a clean improvement - eliminating the manual type map makes the API much simpler. I particularly like the use of  DemangleName(). Thanks for doing all this work!

PR #213, opened 2025-12-19
Improve test coverage for Python support

greenc-FNAL:
### Goal

The primary goal of this PR is to significantly improve the test coverage for the Phlex Python plugin and resolve several discovered issues related to memory management, type conversion, and core graph connectivity safety. It also modernizes the build system configuration and coverage tooling.

### Key Changes and Rationale

**1. Introduction of `phlex.Variant` Helper**
- **Changes**: Added a new Python class `Variant` in `plugins/python/python/phlex/__init__.py`.
- **Motivation**: Python algorithms are often generic, but Phlex registration requires specific C++ signatures. Additionally, Python parameter names in a function's `__annotations__` might not match the positional labels defined in the Phlex configuration.
- **Solution**: The `Variant` class allows developers to wrap a callable and associate it with custom annotations and a specific name. This enables the same Python function to be registered multiple times with different types without modifying the original function.
- **Rationale**: This decouples the algorithm implementation from the registration metadata. The C++ bridge was also updated to "unwrap" these objects, retrieving the underlying `phlex_callable` for performance while still using the provided annotations for type mapping.

**2. Robustness Improvements in `modulewrap.cpp`**
- **Reference Counting**: Added missing `Py_DECREF` calls in the error paths of `parse_args`, `md_transform`, and `md_observe`. This prevents Python object leaks when registration fails.
- **Sequence Handling**: Replaced manual sequence checks with `PySequence_Fast`. This allows Phlex to accept any Python sequence (lists, tuples, etc.) for input/output labels.
- **Type Matching & NumPy Fixes**: 
    - Corrected a typo where double-precision NumPy arrays were matched against `double64]]` instead of `float64]]`.
    - Updated type matching logic to use fixed-length substring comparisons, supporting PEP 604 union types (e.g., `ndarray | list`).
- **List-based Type Converters**: Added explicit support for `list[int]`, `list[float]`, and `list[double]`. 
    - **Rationale**: Provides a fallback for users not using NumPy or for cases where standard Python lists are more appropriate.

**3. Core Connectivity & Safety**
- **Changes**: Added null-pointer checks in `phlex/core/edge_maker.cpp` and `phlex/core/edge_maker.hpp`.
- **Motivation**: Misconfigurations in the Jsonnet graph could lead to null ports during edge creation, resulting in segmentation faults.
- **Solution**: The code now throws a `std::runtime_error` with a descriptive message identifying the offending nodes/ports.

**4. Build System & Tooling Modernization**
- **CMake Optimizations**:
[...]

PR #269, opened 2026-01-27
Shift `async_driver` gear into park if exception from workflow graph

knoepfel:
This PR removes a stall that could occur if an exception is triggered in the workflow portion of the framework graph.


knoepfel:
> Does this resolve the remaining segfault issue in #213 or is that still not understood?

Yes, I believe it fixes the segfault issue in #213.  (My local checkout of #213 with this PR's changes applied did not yield any stalled or segfaulting jobs when using `ctest -j24`, the maximum allowed parallelism on the system I was using.)

knoepfel:
@greenc-FNAL, I adjusted the comment and fixed some of the waiting logic...it was more complicated than it needed to be.

greenc-FNAL (approved):

PR #282, opened 2026-02-03
JSON parsing changes from new product query

beojan:
This eliminates need for default constructor.

See #267

### DIFF
--- a/phlex/configuration.cpp
+++ b/phlex/configuration.cpp
@@ -8,8 +8,9 @@
 #include <string_view>
 
 namespace {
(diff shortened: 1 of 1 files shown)


knoepfel (commented):
Changes look reasonable, @beojan.  Can you first rebase the branch on top of the latest commits to `main`?  That should make all of the tests succeed.

knoepfel (approved):
LGTM.  We should make sure the `value_if_exists` function is tested when it is eventually used.  It will not hold up this PR, though

PR #339, opened 2026-02-18
Improve Phlex test coverage

greenc-FNAL:
- **Improve test coverage of Phlex code**
- **Address https://github.com/Framework-R-D/phlex/pull/335#discussion_r2824494351**

### DIFF
--- a/phlex/model/identifier.hpp
+++ b/phlex/model/identifier.hpp
@@ -58,14 +58,6 @@ namespace phlex::experimental {
 }
 
 template <>
-struct fmt::formatter<phlex::experimental::identifier> : formatter<std::string_view> {
-  auto format(phlex::experimental::identifier const& id, format_context& ctx) const
-  {
-    return formatter<std::string_view>::format(std::string_view(id), ctx);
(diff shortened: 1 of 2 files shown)


phlex/model/identifier.hpp
```diff
@@ -57,6 +57,14 @@ namespace phlex::experimental {
   inline std::string_view format_as(identifier const& id) { return std::string_view(id); }
 }
 
+template <>
```
beojan: This shouldn't be necessary. `format_as` is intended to handle this.

test/identifier.cpp
```diff
@@ -65,5 +65,16 @@ int main()
     }
   }
   assert(ok);
+
+  // Additional coverage for identifier edge cases
+  identifier id1("abc");
+  identifier id2("def");
```
beojan: Comment needed about relative ordering, since it isn't alphabetical.

The comparison operators exist purely in order to make it possible to put identifiers into a `map` or `set`, not because they are inherently ordered.

PR #344, opened 2026-02-19
Actually using the new product_query (Part 1)

beojan:
Mostly switching to using `identifier` members, along with adding the ability to resolve products without a suffix.


phlex/model/identifier.hpp
```diff
@@ -50,7 +62,7 @@ namespace phlex::experimental {
 
   private:
     std::string content_;
-    std::uint64_t hash_;
+    std::uint64_t hash_{0};
```
knoepfel: Is `0` the calculated hash of the empty string?

beojan: No, it isn't. This is only here so I can ` = default` the default constructor without risking undefined behaviour. And the default constructor is only there so other classes that contain identifiers can be default constructed.

phlex/model/product_store.cpp
```diff
@@ -34,4 +36,11 @@ namespace phlex::experimental {
     }
     return b;
   }
+
+  algorithm_name product_store::default_source()
+  {
+    using namespace literals;
```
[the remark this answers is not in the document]
beojan: :+1:

knoepfel:
@beojan, this looks good.  Some conflicts need some resolution but then we should be good to go.

PR #349, opened 2026-02-22
refactor: replace technology if-else chain with C++20 lookup table

wwuoneway:
**Summary**： 
This is a **refactor and clean-up test** to modernize the module registration logic.

**Why**
- Cleaner Code: Replaced a verbose if-else block with a declarative std::unordered_map.
- Modern Standards: Leveraged the project's C++20 configuration to use std::string_view for efficient, zero-allocation lookups.
- Scalability: Makes it easier to add or modify supported technologies in the future without nesting more logic.

### DIFF
--- a/form/form_module.cpp
+++ b/form/form_module.cpp
@@ -119,13 +119,11 @@ PHLEX_REGISTER_ALGORITHMS(m, config)
   std::cout << "  output_file: " << output_file << "\n";
   std::cout << "  technology: " << tech_string << "\n";
 
-  // Using a static map for O(1) lookup
-  static std::unordered_map<std::string_view, int> const tech_lookup = {
+  std::unordered_map<std::string_view, int> const tech_lookup = {
     {"ROOT_TTREE", form::technology::ROOT_TTREE},
     {"ROOT_RNTUPLE", form::technology::ROOT_RNTUPLE},
     {"HDF5", form::technology::HDF5}};
 
-  // C++20 allows finding a string_view key efficiently
   auto it = tech_lookup.find(tech_string);
 
   if (it == tech_lookup.end()) {


form/form_module.cpp
```diff
@@ -74,6 +78,8 @@ namespace {
       for (auto const& [product_name, product_ptr] : store) {
         // product_name: "tracks" (from the map key)
         // product_ptr: pointer to the actual product data
+        if (!product_ptr)
+          continue;
```
gemmeren: Is this an expected outcome or something that we may want to warn about?

form/form_module.cpp
```diff
@@ -107,25 +113,27 @@ PHLEX_REGISTER_ALGORITHMS(m, config)
 
   // Extract configuration from Phlex config
   std::string output_file = config.get<std::string>("output_file", "output.root");
```
gemmeren: While you are at it, make this const as well?

form/form_module.cpp
```diff
@@ -106,26 +112,28 @@ PHLEX_REGISTER_ALGORITHMS(m, config)
...
+    {"HDF5", form::technology::HDF5}};
 
-  if (tech_string == "ROOT_TTREE") {
-    technology = form::technology::ROOT_TTREE;
-  } else if (tech_string == "ROOT_RNTUPLE") {
-    technology = form::technology::ROOT_RNTUPLE;
-  } else if (tech_string == "HDF5") {
-    technology = form::technology::HDF5;
-  } else {
+  // C++20 allows finding a string_view key efficiently
```
knoepfel: This comment about C++20 is true, but it's not necessary in the code.  I suggest removing it.

PR #379, opened 2026-03-03
Add FORM-based source provider

barnaliy:
Adds form_source.cpp implementing a FORM-based provider that retrieves data products from FORM storage into Phlex.
This is the source-side complement to form_module.cpp.

### DIFF
--- a/form/form_source.cpp
+++ b/form/form_source.cpp
@@ -1,7 +1,7 @@
-#include "phlex/source.hpp"
 #include "form/config.hpp"
 #include "form/form.hpp"
 #include "form/technology.hpp"
+#include "phlex/source.hpp"
 
 #include <memory>
 #include <stdexcept>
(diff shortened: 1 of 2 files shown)


form/form_source.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+  // --- Create shared form_interface ---
+  auto reader = std::make_shared<form::experimental::form_interface>(input_cfg, tech_cfg);
+
+  // --- Register providers ---
+  // FIXME: Prototype 0.1 -- product names and types hardcoded.
+  // product_query fields must be compile-time _id literals, not runtime strings.
+  // To add new products, add new s.provide() blocks below following the same pattern.
+
+  s.provide("provide_i",
```
gemmeren: Will this be needed for every data product we read from input? If so, we may want to think about a way to shorten or auto-generate these functions.

barnaliy: This is already noted in the FIXME comment — it's a known limitation of Prototype 0.1 that will be addressed as the type list grows.

form/form_source.cpp
```diff
@@ -0,0 +1,83 @@
...
+using namespace phlex::experimental::literals;
+
+PHLEX_REGISTER_PROVIDERS(s, config)
+{
+  std::cout << "Registering FORM source providers...\n";
+
+  // --- Extract configuration ---
+  std::string const input_file = config.get<std::string>("input_file");
+  std::string const creator = config.get<std::string>("creator");
+  std::string const tech_string = config.get<std::string>("technology", "ROOT_TTREE");
```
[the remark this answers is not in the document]
gemmeren: OK for this PR and the first version, but maybe we can create a git issue?

form/form_source.cpp
gemmeren: Here too I wouldn't mind a git issue

wwuoneway (commented):
I have a structural concern regarding the separation between `form_module` and `form_source`, as both currently communicate directly with Phlex. `form_source` appears to focus on reading input files, while `form_module` likely handles the integration with Phlex. Should `form_source` be a standalone module that communicates directly with Phlex, or would it be more appropriate as an internal component used by `form_module`?

form/form_source.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+  s.provide("provide_i",
+            [reader, creator](data_cell_index const& id) -> int {
+              std::cout << "\n=== form_source: providing i ===\n";
+              form::experimental::product_with_name pb{"i", nullptr, &typeid(int)};
+              reader->read(creator, id.to_string(), pb);
+              if (!pb.data) {
+                throw std::runtime_error("FORM read returned null data for product: i");
+              }
+              return *static_cast<int const*>(pb.data);
```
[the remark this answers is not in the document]
pcanal: However, you actually need both in case the think is read but never past on to Phlex.

wwuoneway:
If the current version passes all the checks, I'm fine with going with it for now and following up on the known issues in a later update.

wwuoneway:
Closing this PR as it has been superseded by #552, which includes theses changes, revises the implementation, and resolves the conflicts with the latest main branch.

PR #449, opened 2026-03-24
Avoid fixed-size `long` integral types for macOS

knoepfel:
As mentioned in a Slack discussion:

> The fixed-width integer type `std::int64_t` is a typedef (https://en.cppreference.com/w/cpp/types/integer.html).   In particular, `std::int64_t` on macOS is a typedef to `long long`.  So there is a mismatch between the `long` that the Python system expects, and the `long long` that is calculated by the `type_id` system.

This PR is to work around the problem until we have consistent treatment of types between C++ and Python.


greenc-FNAL:

wlav (approved):
As discussed, fine for now. I'll reconsider when dealing with the refactoring and/or pull in the converters from a different place (pybind11 or cppyy) once we go to real data products.

PR #495, opened 2026-04-08
`clang-tidy`: resolve `bugprone-unchecked-optional-access`

greenc-FNAL:
- `test/form/form_basics_test.cpp`
- `test/identifier.cpp`

`clang-tidy` does not recognize Catch2's `REQUIRE` macro as protecting
the target `std::optional` against non-existent-value access; use `//
NOLINT` to silence the warning.

- `plugins/python/src/modulewrap.cpp`

`clang-tidy` correctly noted that there was a (theoretical) path to
unprotected access to an `std::optional`. We now throw an
`std::logic_error` and protect it from coverage validation as a `should
never happen` untestable case.


plugins/python/src/modulewrap.cpp
```diff
@@ -1186,6 +1186,11 @@ static PyObject* sc_provide(py_phlex_source* src, PyObject* args, PyObject* kwds
     std::string msg;
     if (msg_from_py_error(msg, false)) {
       throw std::runtime_error("output specification error: " + msg);
+    } else {
+      // LCOV_EXCL_START
+      // no error message? should never happen, but just in case
+      throw std::logic_error("output specification error: unknown query validation error");
+      // LCOV_EXCL_STOP
```
knoepfel: Instead of this, can you use [`std::unreachable()`](https://en.cppreference.com/w/cpp/utility/unreachable.html)?

greenc-FNAL: It's not that the code is actually unreachable: it's that it's vanishingly unlikely to happen (there'd have to be an issue with the query validation that prevented a Python error being generated, which should be tested for there), and it's not testable here. `std::unreachable` would cause the `throw` to be optimized away in release builds, allowing possible unchecked access to the `std::optional` below or other unwanted behavior.

greenc-FNAL: See 8ccdc78e—@wlav, would you be able to comment on the following extract from the commit message?

> Additionally, the claim is that the `std::runtime_error()` throw will never be caught, and we should probably be using `Python_SetErr()` here and returning `nullptr` instead of throwing exceptions.

plugins/python/src/modulewrap.cpp
```diff
@@ -1182,11 +1182,13 @@ static PyObject* sc_provide(py_phlex_source* src, PyObject* args, PyObject* kwds
   // translate and validate the output query
   auto opq = validate_query(output);
   if (!opq.has_value()) {
-    // validate_query will have set a python exception
+    // LCOV_EXCL_START
+    // validate_query _will_ have set a python exception with details about the
+    //  error, so just propagate it as a C++ exception
     std::string msg;
```
knoepfel: Might initialize `msg` to something like "(unknown)" so that if `msg_from_py_error` returns `false`, the user doesn't see an exception message like: `"output specification error:"` with nothing following the colon.

knoepfel (approved):

PR #497, opened 2026-04-09
`clang-tidy`: resolve `bugprone-narrowing-conversions`

greenc-FNAL:


plugins/python/src/configwrap.cpp
```diff
@@ -77,48 +77,57 @@ static PyObject* pcm_subscript(py_config_map* pycmap, PyObject* pykey)
     if (k.second /* is array */) {
       if (k.first == boost::json::kind::bool_) {
         auto const& cvalue = pycmap->ph_config->get<std::vector<bool>>(ckey);
-        pyvalue = PyTuple_New(cvalue.size());
-        for (Py_ssize_t i = 0; i < (Py_ssize_t)cvalue.size(); ++i) {
+        auto const cvalue_size = (Py_ssize_t)cvalue.size();
+        pyvalue = PyTuple_New(cvalue_size);
+        for (Py_ssize_t i = 0; i < cvalue_size; ++i) {
           PyObject* item = PyLong_FromLong((long)cvalue[i]);
```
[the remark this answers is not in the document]
greenc-FNAL: @copilot apply changes based on [this feedback](https://github.com/Framework-R-D/phlex/pull/497#discussion_r3059045597).

Specifically, please ensure that definitions of `cvalue_size` use `static_cast<Py_ssize_t>()` instead of C-style casts, that the value does not suffer a truncating conversion, and if it does, that a suitable Python error and return takes place.

greenc-FNAL:
@copilot Please add/enhance unit testing for [plugins/python/src/configwrap.cpp](https://app.codecov.io/gh/Framework-R-D/phlex/pull/497?src=pr&el=tree&filepath=plugins%2Fpython%2Fsrc%2Fconfigwrap.cpp&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Framework-R-D#diff-cGx1Z2lucy9weXRob24vc3JjL2NvbmZpZ3dyYXAuY3Bw) to address missing test coverage for this PR

phlex/utilities/resource_usage.cpp
```diff
@@ -24,7 +24,8 @@ namespace {
     rusage used;
     getrusage(RUSAGE_SELF, &used);
     auto const [secs, microsecs] = used.ru_utime;
-    return {.elapsed_time = secs + microsecs / 1e6, .max_rss = used.ru_maxrss / mem_denominator};
+    return {.elapsed_time = double(secs) + double(microsecs) / 1000000.0,
```
knoepfel: The expression `1e6` is of type `double` and is easier to read than `1000000.0`.  Suggest reverting that part.

phlex/utilities/resource_usage.cpp
```diff
@@ -39,7 +40,8 @@ namespace phlex::experimental {
     auto const [elapsed_time, max_rss] = get_rusage();
     auto const end_wall = steady_clock::now();
     auto const cpu_time = elapsed_time - begin_cpu_;
-    auto const real_time = duration_cast<nanoseconds>(end_wall - begin_wall_).count() / 1e9;
+    auto const real_time =
+      double(duration_cast<nanoseconds>(end_wall - begin_wall_).count() / 1000000000);
```
knoepfel: This calculation is incorrect.  The `duration_cast<nanoseconds>(...).count()` yields an integer value, which is then divided by one billion, thus yielding granularity at the level of a second, which is too coarse.  If anything, the numerator can be explicitly cast to a `double` (the denominator of `1e9` is already of type `double`, and it is easier to read than `1000000000`).

test/form/writer.cpp
```diff
@@ -27,7 +27,7 @@ void generate(std::vector<float>& vrand, int size)
   for (int nelement = 0; nelement < npx; ++nelement) {
     int rand1 = rand() % 32768; // NOLINT(concurrency-mt-unsafe) - Single-threaded test
     int rand2 = rand() % 32768; // NOLINT(concurrency-mt-unsafe) - Single-threaded test
-    float random = float(rand1 * 32768 + rand2) / (32768 * 32768);
+    float random = static_cast<float>(rand1 * 32768 + rand2) / (32768.0f * 32768.0f);
```
knoepfel: I'd keep the denominator of `32768 * 32768`—we shouldn't need to specify them as floats.

plugins/python/src/modulewrap.cpp
```diff
@@ -777,7 +777,12 @@ static bool insert_input_converters(py_phlex_module* mod,
       // TODO: these are hard-coded std::vector <-> numpy array mappings, which is
       // way too simplistic for real use. It only exists for demonstration purposes,
       // until we have an IDL
-      std::string_view dtype{inp_type.begin() + inp_type.rfind('['), inp_type.end()};
+      auto const dtype_pos = inp_type.rfind('[');
+      if (dtype_pos == std::string::npos) {
+        PyErr_Format(PyExc_TypeError, "unsupported collection input type \"%s\"", inp_type.c_str());
+        return false;
+      }
+      std::string_view dtype = std::string_view{inp_type}.substr(dtype_pos);
```
knoepfel: This pattern appears multiple times in this file.  Perhaps it can be consolidated in a small free function that  is invoked so that (e.g.):

```cpp
std::string_view collection_type(std::string const& out_type)
{
    auto const dtype_pos = out_type.rfind('[');
    if (dtype_pos == std::string::npos) {
      PyErr_Format(PyExc_TypeError, "unsupported collection output type \"%s\"", out_type.c_str());
      return ""sv;
    }
    std::string_view dtype{out_type};
    return dtype.substr(dtype_pos);
}

// When used...
auto const dtype = collection_type(out_type);
if (dtype.empty()) {
  return false;
}
...
```

plugins/python/src/configwrap.cpp
```diff
@@ -77,48 +77,81 @@ static PyObject* pcm_subscript(py_config_map* pycmap, PyObject* pykey)
...
         auto const& cvalue = pycmap->ph_config->get<std::vector<bool>>(ckey);
-        pyvalue = PyTuple_New(cvalue.size());
-        for (Py_ssize_t i = 0; i < (Py_ssize_t)cvalue.size(); ++i) {
+        if (cvalue.size() > static_cast<std::size_t>(PY_SSIZE_T_MAX)) {
+          PyErr_Format(PyExc_OverflowError, "array is too large to convert to a Python tuple");
+          return nullptr;
+        }
+        auto const cvalue_size = static_cast<Py_ssize_t>(cvalue.size());
+        pyvalue = PyTuple_New(cvalue_size);
+        for (Py_ssize_t i = 0; i < cvalue_size; ++i) {
```
knoepfel: This pattern is repeated multiple times in the file.  Can it be consolidated into a single free function that is invoked where needed?

test/python/source.cpp
```diff
@@ -6,29 +6,31 @@ using namespace phlex;
...
     .output_product(product_query{.creator = "input", .layer = "event", .suffix = "k"});
 
-  s.provide("provide_f1",
-            [](data_cell_index const& id) -> float { return (float)((id.number() % 100) / 100.0); })
-    .output_product(product_query{.creator = "input", .layer = "event", .suffix = "f1"});
   s.provide(
-     "provide_f2",
-     [](data_cell_index const& id) -> float { return 1.0f - (float)((id.number() % 100) / 100.0); })
+     "provide_f1",
+     [](data_cell_index const& id) -> float { return (float)(((int)id.number() % 100) / 100.0); })
```
knoepfel: Oof...a lot of hard-casting here (and elsewhere in this file).  Is it possible to clean this up?

knoepfel (commented):
@greenc-FNAL, some comments for your consideration below.

greenc-FNAL:
@copilot Please address the review comments at https://github.com/Framework-R-D/phlex/pull/497#pullrequestreview-4099130676, paying special attention to not introducing new `clang-tidy` check issues with any changes.

greenc-FNAL:
@Phlexbot clang-fix

test/form/writer.cpp
greenc-FNAL: Note: at least the first `32768f` is necessary to avoid:

```console
[{
	"resource": "/workspaces/phlex/test/form/writer.cpp",
	"owner": "clang-tidy",
	"severity": 4,
	"message": "narrowing conversion from constant value 1073741824 of type 'int' to 'float' [bugprone-narrowing-conversions]",
	"startLineNumber": 30,
	"startColumn": 64,
	"endLineNumber": 30,
	"endColumn": 64
}]
```

knoepfel: Ah, I see.  Didn't realize `int` to `float` was narrowing.  I defer to your judgment

greenc-FNAL: It _can_ be narrowing. In this case (with 2^30), you only have 23 bits of mantissa, so you will lose some precision. If we're serious about resolving these issues, we'll need an `f` modifier there.

knoepfel (approved):

PR #500, opened 2026-04-09
`clang-tidy`: resolve `bugprone-macro-parentheses`

greenc-FNAL:
Warnings were confined to two header files and determined to be false
positives—parentheses around uses of parameters in C++ code generation
macros are not only superfluous, but disastrous for successful
compilation.

Macros are now protected from this check _en bloc_ due to the
impossibility of threading perl-line `NOLINT` statements in macro
definitions with line continuations.


wlav (dismissed):
LGTM

knoepfel (approved):

PR #501, opened 2026-04-09
Add full_product_spec and use that in provider API

beojan:
For now product_query has a conversion operator so the tests don't have to be rewritten. This conversion operator has a commented out `deprecated` annotation (because we use `-Werror`).

### DIFF
--- a/phlex/core/product_query.hpp
+++ b/phlex/core/product_query.hpp
@@ -3,6 +3,7 @@
 
 #include "phlex/phlex_core_export.hpp"
 
+#include "phlex/model/full_product_spec.hpp"
 #include "phlex/model/identifier.hpp"
 #include "phlex/model/product_specification.hpp"
 #include "phlex/model/product_store.hpp"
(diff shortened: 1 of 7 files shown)


phlex/core/product_query.hpp
```diff
@@ -80,10 +81,28 @@ namespace phlex {
...
+
     std::string to_string() const;
 
     bool operator==(product_query const& rhs) const;
     std::strong_ordering operator<=>(product_query const& rhs) const;
+
+    // Transitional automatic conversion operator so I don't have to rewrite all the tests
+    // The deprecated annotation is commented out because we have -Werror on
+    // [[deprecated(
+    //   "Generation of a full_product_spec from a product_query is only transitionally supported")]]
```
knoepfel, line 95: The deprecation attribute is fine for our users.  They decide how to build their own Phlex-dependent code, with or without `-Werror` and `-Wno-error=deprecated-declarations`.

For our own purposes, we'll continue building with `-Werror` and not exclude deprecated declarations.  We should convert all tests to use the encouraged API.  If we want to test the deprecation, we can add a dedicated compile-only test that verifies the deprecation message.  I'm not sure that's necessary, though.

phlex/model/full_product_spec.hpp
```diff
@@ -0,0 +1,54 @@
...
+#include "phlex/model/product_specification.hpp"
+#include "phlex/model/type_id.hpp"
+
+#include "boost/container_hash/hash.hpp"
+#include "fmt/format.h"
+
+namespace phlex::experimental {
+  class PHLEX_MODEL_EXPORT full_product_spec {
+  public:
+    full_product_spec(product_specification spec, identifier layer, identifier stage) :
```
knoepfel, line 16: I am leery of creating a full product specification when what is constructed is not, in fact, a full specification (the type is not set until later).  We did this for `product_query`/`product_specification`, and I'd like to fix this...specifically, I'd like to avoid the need for setters wherever possible.

I think this means that the provider's `output_product` registration interface requires all of the data-product fields encapsulated by `product_specification`, layer, and stage, but maybe they can be received from the user in a more ergonomic way (e.g., require the arguments `algorithm_name`, suffix, layer, stage, where stage can default to an identifier that denotes the current process):

```.cpp
m.provide(...)
 .output_product(
   /* algorithm_name*/ creator,  
   /* identifier */ suffix, 
   /* identifier */ layer, 
   /* identifier, possibly defaulted */ stage
  )
```

How these fields are packaged is then an implementation detail.

knoepfel (commented):
Thanks, @beojan, for starting this.  See below for some comments.

beojan:
@copilot resolve the merge conflicts in this pull request

phlex/model/full_product_spec.hpp
beojan: #602 takes this approach (drafted for now). I'll close this PR.

PR #525, opened 2026-04-13
improve error handling consistency

wlav:
Originally noticed by @greenc-FNAL, see here: https://github.com/Framework-R-D/phlex/pull/495#discussion_r3054542041, there are cases in the current code where a C++ exception is thrown where to code can still be reasonably be expected to return to Python for error handling. This PR removes two cases of throwing C++ exceptions in favor of setting Python exceptions and returning `nullptr`.

### DIFF
--- a/plugins/python/src/modulewrap.cpp
+++ b/plugins/python/src/modulewrap.cpp
@@ -345,28 +345,24 @@ namespace {
         Py_DECREF(phlexmod);
       }
 
-      if (!normalizer) {
-        std::string msg;
-        if (msg_from_py_error(msg, false))
-          throw std::runtime_error("unable to retrieve the phlex type normalizer: " + msg);
-      }
+      if (!normalizer)
+        return "";
     }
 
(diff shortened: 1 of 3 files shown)


greenc-FNAL (commented):
LGTM—I assume `return ""` _vs_ `return nullptr` was a conscious choice?

I asked Copilot to try to improve the test coverage for the flagged lines missing coverage here, and it came up with https://github.com/Framework-R-D/phlex/pull/530/changes/8c288881ce4af7537b29056a2f31f30ce81c0c58: feel free to cherry-pick, take inspiration from, or ignore as you prefer.

wlav:
> LGTM—I assume `return ""` _vs_ `return nullptr` was a conscious choice?

Yes; `nullptr` was never on the table, but I was thinking of going for `std::optional` there, then was afraid of triggering more coverage issues, so figured that since `""` (no annotation) is always an error anyway and since this is not a public function, might as well deal with it this way.

greenc-FNAL (approved):

greenc-FNAL (approved):

PR #531, opened 2026-04-14
Add branch push triggers for `release/*` branches

greenc-FNAL:
- Also remove references to unused `develop` and `development` in workflows and documentation.

### DIFF
--- a/.github/workflows/actionlint-check.yaml
+++ b/.github/workflows/actionlint-check.yaml
@@ -7,6 +7,8 @@ permissions:
 
 "on":
   pull_request:
+  push:
+    branches: ["releases/*"]
   workflow_dispatch:
     inputs:
       ref:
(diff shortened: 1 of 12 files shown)


knoepfel (commented):
Is anyone making pushes directly to the `release/*` branches?  We did early on for `v0.1.0`, but I think we've only used pull requests since then.  If no direct pushes are being made, do these changes still need to be implemented?

knoepfel:
> Is anyone making pushes directly to the `release/*` branches? We did early on for `v0.1.0`, but I think we've only used pull requests since then. If no direct pushes are being made, do these changes still need to be implemented?

Talked with @greenc-FNAL.  Triggering on pushes also includes triggering on merges.

knoepfel (approved):

PR #533, opened 2026-04-16
Test that repeaters never emit data products for flush counts of 0

knoepfel:
This PR tests a behavior that is required for multilayer algorithms where the higher-layer data cell exists, but the lower-layer data cell does not exist (e.g., because the product was filtered out at the lower layer). The higher-layer data cell thus has a flush count of 0 (i.e., there are no lower-layer data-cells that correspond to it).  In such a case, we do not want the higher-layer data product to be sent to a multilayer join node, which would wait indefinitely for the lower-layer data product that will never arrive.

### DIFF
--- a/test/repeater_node_test.cpp
+++ b/test/repeater_node_test.cpp
@@ -329,3 +329,28 @@ TEST_CASE("Test warning message if there are cached messages", "[multithreading]
     CHECK_THAT(warning, Catch::Matchers::ContainsSubstring("Product not yet received"));
   }
 }
+
+TEST_CASE("Test data-only with flush count zero", "[multithreading]")
+{
+  // This behavior is required for multilayer algorithms where the higher-layer data cell exists,
+  // but the lower-layer data cell does not exist (e.g. because the product was filtered out at the
+  // lower layer). We do not want the higher-layer data product to be sent to a multilayer join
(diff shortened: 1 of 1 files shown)


marcpaterno (approved):

marcpaterno:
I agree this test does what it is intended to do. But I'm uneasy with the interaction of filtering with other parts of the system, and would like to have a discussion about the interactions and what we want filtering to look like.

Wanting that discussion should not delay the merging of this PR.

knoepfel:
> I agree this test does what it is intended to do. But I'm uneasy with the interaction of filtering with other parts of the system, and would like to have a discussion about the interactions and what we want filtering to look like.
> 
> Wanting that discussion should not delay the merging of this PR.

Filtering, in this context, means that maybe a subrun has been written to disk, but none of the spills associated with that subrun have been written to disk because there was nothing interesting in them.  In a subsequent job, an algorithm may want to process a data product from the subrun and the spill.  That means that algorithm will have two repeaters associated with it (one for the subrun, and one for the spill).  However, because there are no spills for the particular subrun in the file from the previous process, the algorithm cannot execute...even if the subrun has a data product that can serve as input to the algorithm.

We can chat more later.

PR #541, opened 2026-04-22
Fix real heap leaks in py_callback registrations; suppress Boost false positive

greenc-FNAL:
plugins/python/src/modulewrap.cpp
- In md_transform() and md_observe(), six heap allocations of the form
  'auto* pyc = new py_callback_N{callable}; register(*pyc, ...)' were
  real memory leaks.  The registration methods (transform(), observe())
  accept their callable argument by value: graph_proxy<T>::observe()
  passes it by value into glue<T>::observe(), which constructs an
  algorithm_bits wrapping std::function via delegate(), and ultimately
  stores a copy of the callable object inside the TBB flow-graph node.
  The original heap-allocated py_callback_N object was therefore never
  referenced again after registration and was never freed.

  Since py_callback's destructor is intentionally a no-op (Python
  reference-count cleanup is deferred to the Phlex shutdown hook; see
  the ~py_callback() comment), constructing a temporary and passing it
  directly is safe and avoids the allocation entirely:

    Before: auto* pyc = new py_callback_1{callable};
            mod->ph_module->transform(pyname, *pyc, concurrency::serial)
    After:  mod->ph_module->transform(pyname, py_callback_1{callable}, concurrency::serial)

  This also removes the pre-existing 'TODO: leaks, but has program
  lifetime' comment in the case-1 branch of md_transform(); that
  reasoning was incorrect because the callable is copied into the graph
  at registration time, so the raw pointer was always orphaned.

phlex/app/load_module.cpp
- clang-analyzer-cplusplus.NewDeleteLeaks and .NewDelete reported three
  warnings at the call to plugin_loader<driver_creator_t>() in
  load_driver().  The trace leads into boost::split() and ultimately
  into Boost.StringAlgo's is_any_ofF<char> copy constructor
  (classification.hpp), which uses 'new' internally and manages the
  allocation via its own reference-counting scheme.  The analyser cannot
  follow that ownership chain through the Boost internals and
  misidentifies the allocation as leaked.  The contradictory pair of
  'potential leak' and 'attempt to release already released memory' for
  the same object at the same site confirms the report is spurious.

  Suppressed with NOLINTNEXTLINE and an explanatory comment.

### DIFF
--- a/docs/dev/clang-tidy-fixes-2026-04.md
+++ b/docs/dev/clang-tidy-fixes-2026-04.md
@@ -40,8 +40,10 @@
   - [PR #538](https://github.com/Framework-R-D/phlex/pull/538)
 - [x] [clang-analyzer-core.NullDereference](https://clang.llvm.org/extra/clang-tidy/checks/clang-analyzer/core.NullDereference.html) (1)
   - [PR #538](https://github.com/Framework-R-D/phlex/pull/538)
-- [ ] [clang-analyzer-cplusplus.NewDelete](https://clang.llvm.org/extra/clang-tidy/checks/clang-analyzer/cplusplus.NewDelete.html) (1)
-- [ ] [clang-analyzer-cplusplus.NewDeleteLeaks](https://clang.llvm.org/extra/clang-tidy/checks/clang-analyzer/cplusplus.NewDeleteLeaks.html) (4)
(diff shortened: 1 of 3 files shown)


wlav:
Yes, I already have similar changes in my numba-support branch.

knoepfel (approved):

PR #544, opened 2026-04-22
`clang-tidy`: resolve `cppcoreguidelines-avoid-non-const-global-variables`

greenc-FNAL:
- Audit all occurrences
- Fix unnecessary occurrence in `test/type_deduction.cpp`
- Add justification comments and ignore via `NOLINT*` annotations


form/form_module.cpp
```diff
@@ -108,6 +108,8 @@ namespace {
 
 }
 
+// BOOST_DLL_ALIAS creates a non-const exported function pointer; required by the dynamic linker.
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
 PHLEX_REGISTER_ALGORITHMS(m, config)
```
knoepfel: Can we find a way to suppress the warning without needing to place the comment immediately before our public interface?

greenc-FNAL: The only thing I can think of to disable the diagnostic is to have CMake mark the `plugin_macros.hpp` header as a system header.
`//` and `/* ... */` comments with `NOLINT` get CPP'd away too early, and clang-tidy doesn't support `_Pragma` instructions like the compiler itself does.

knoepfel: If the warning is being triggered by something in Boost, can't we instead mark one of the Boost headers as a system header?

knoepfel (approved):

PR #546, opened 2026-04-23
`clang-tidy`: resolve `cppcoreguidelines-init-variables`

greenc-FNAL:
- All justified as being initialized either unconditionally via reference, `operator>>()`, or conditionally but comprehensively via `if`/`else`.


greenc-FNAL:
@aolivier23 @gemmeren @wwuoneway The [Copilot review](https://github.com/Framework-R-D/phlex/pull/546#pullrequestreview-4163945100) dings all uninitialized definitions handled here, stating that they should be initialized for robustness reasons. Please advise on whether I should do this or preserve and annotate the original code.

greenc-FNAL:
> @aolivier23 @gemmeren @wwuoneway The [Copilot review](https://github.com/Framework-R-D/phlex/pull/546#pullrequestreview-4163945100) dings all uninitialized definitions handled here, stating that they should be initialized for robustness reasons. Please advise on whether I should do this or preserve and annotate the original code.

Per @knoepfel suggestion at the Phlex developers meeting today, I intend to initialize explicitly and allow the optimizing compiler to elide if/when appropriate unless you express a different preference.

knoepfel (approved):

PR #551, opened 2026-04-24
`clang-tidy`: resolve `cppcoreguidelines-owning-memory`

greenc-FNAL:
- **Follow C++ Core Guidelines for owning pointers when demangling symbols**
- **Follow C++ Core Guidelines in `test/form/reader.cpp`**
- **Avoid unnecessary (and likely problematic) `new()` in `doRead()`**
- **Clarify and simplify `getIndex()` while avoiding raw `delete`**
- **Qualify linkage as explicity `PUBLIC`**
- **Avoid unnecessary (and likely problematic) `new()` operations**
- **Use `unique_ptr` with a custom deleter for clarity and consistency**
- **Follow C++ Core Guidelines for owning pointers**

### DIFF
--- a/form/root_storage/root_tbranch_read_container.cpp
+++ b/form/root_storage/root_tbranch_read_container.cpp
@@ -61,7 +61,7 @@ bool ROOT_TBranch_Read_ContainerImp::read(int id, void const** data, std::type_i
   if (dictInfo->Property() & EProperty::kIsFundamental) {
     // NOLINTNEXTLINE(cppcoreguidelines-pro-type-static-cast-downcast)
     auto fundInfo = static_cast<TDataType*>(dictInfo); // Already checked to be fundamental
-    branchBuffer = new char[fundInfo->Size()];
+    branchBuffer = nullptr;
     branchStatus = m_tree->SetBranchAddress(col_name().c_str(),
                                             reinterpret_cast<void*>(&branchBuffer),
                                             nullptr,
@@ -74,7 +74,7 @@ bool ROOT_TBranch_Read_ContainerImp::read(int id, void const** data, std::type_i
                                " (col_name='" + col_name() + "', type='" + DemangleName(type) +
                                "')");
     }
-    branchBuffer = klass->New();
+    branchBuffer = nullptr;
     branchStatus = m_tree->SetBranchAddress(
       col_name().c_str(), reinterpret_cast<void*>(&branchBuffer), klass, EDataType::kOther_t, true);
   }
--- a/form/root_storage/root_tbranch_read_container.cpp
+++ b/form/root_storage/root_tbranch_read_container.cpp
@@ -63,43 +63,43 @@ bool ROOT_TBranch_Read_ContainerImp::read(int id, void const** data, std::type_i
     auto fundInfo = static_cast<TDataType*>(dictInfo); // Already checked to be fundamental
     switch (fundInfo->GetType()) {
     case kChar_t:
-      branchBuffer = new char;
+      branchBuffer = new Char_t;
       break;
     case kUChar_t:
-      branchBuffer = new unsigned char;
+      branchBuffer = new UChar_t;
       break;
     case kShort_t:
-      branchBuffer = new short;
+      branchBuffer = new Short_t;
       break;
     case kUShort_t:
-      branchBuffer = new unsigned short;
+      branchBuffer = new UShort_t;
       break;
     case kInt_t:
-      branchBuffer = new int;
+      branchBuffer = new Int_t;
       break;
     case kUInt_t:
-      branchBuffer = new unsigned int;
+      branchBuffer = new UInt_t;
       break;
     case kLong_t:
-      branchBuffer = new long;
+      branchBuffer = new Long_t;
       break;
     case kULong_t:
-      branchBuffer = new unsigned long;
+      branchBuffer = new ULong_t;
       break;
     case kLong64_t:
-      branchBuffer = new int64_t;
+      branchBuffer = new Long64_t;
       break;
     case kULong64_t:
-      branchBuffer = new uint64_t;
+      branchBuffer = new ULong64_t;
       break;
     case kFloat_t:
-      branchBuffer = new float;
+      branchBuffer = new Float_t;
       break;
     case kDouble_t:
-      branchBuffer = new double;
+      branchBuffer = new Double_t;
       break;
     case kBool_t:
-      branchBuffer = new bool;
+      branchBuffer = new Bool_t;
       break;
     default:
       throw std::runtime_error(
--- a/form/root_storage/root_tbranch_read_container.cpp
+++ b/form/root_storage/root_tbranch_read_container.cpp
@@ -9,6 +9,8 @@
 #include "TLeaf.h"
 #include "TTree.h"
 
+#include <gsl/pointers>
+
 #include <unordered_map>
 
 using namespace form::detail::experimental;
@@ -49,7 +51,7 @@ bool ROOT_TBranch_Read_ContainerImp::read(int id, void const** data, std::type_i
   if (id > m_tree->GetEntries())
     return false;
 
-  void* branchBuffer = nullptr;
+  gsl::owner<void*> branchBuffer = nullptr;
   auto dictInfo = TDictionary::GetDictionary(type);
   int branchStatus = 0;
 
@@ -106,11 +108,8 @@ bool ROOT_TBranch_Read_ContainerImp::read(int id, void const** data, std::type_i
         std::string{"ROOT_TBranch_ContainerImp::read unsupported fundamental type: "} +
         DemangleName(type));
     };
-    branchStatus = m_tree->SetBranchAddress(col_name().c_str(),
-                                            reinterpret_cast<void*>(branchBuffer),
-                                            nullptr,
-                                            EDataType(fundInfo->GetType()),
-                                            false);
+    branchStatus = m_tree->SetBranchAddress(
+      col_name().c_str(), branchBuffer, nullptr, EDataType(fundInfo->GetType()), false);
   } else {
     auto klass = TClass::GetClass(type);
     if (!klass) {
@@ -118,7 +117,7 @@ bool ROOT_TBranch_Read_ContainerImp::read(int id, void const** data, std::type_i
                                " (col_name='" + col_name() + "', type='" + DemangleName(type) +
                                "')");
     }
-    branchBuffer = klass->New();
+    branchBuffer = gsl::owner<void*>(klass->New());
     branchStatus = m_tree->SetBranchAddress(
       col_name().c_str(), reinterpret_cast<void*>(&branchBuffer), klass, EDataType::kOther_t, true);
   }
--- a/form/root_storage/root_tbranch_read_container.cpp
+++ b/form/root_storage/root_tbranch_read_container.cpp
@@ -61,12 +61,56 @@ bool ROOT_TBranch_Read_ContainerImp::read(int id, void const** data, std::type_i
   if (dictInfo->Property() & EProperty::kIsFundamental) {
     // NOLINTNEXTLINE(cppcoreguidelines-pro-type-static-cast-downcast)
     auto fundInfo = static_cast<TDataType*>(dictInfo); // Already checked to be fundamental
-    branchBuffer = nullptr;
+    switch (fundInfo->GetType()) {
+    case kChar_t:
+      branchBuffer = new char;
+      break;
+    case kUChar_t:
+      branchBuffer = new unsigned char;
+      break;
+    case kShort_t:
+      branchBuffer = new short;
+      break;
+    case kUShort_t:
+      branchBuffer = new unsigned short;
+      break;
+    case kInt_t:
+      branchBuffer = new int;
+      break;
+    case kUInt_t:
+      branchBuffer = new unsigned int;
+      break;
+    case kLong_t:
+      branchBuffer = new long;
+      break;
+    case kULong_t:
+      branchBuffer = new unsigned long;
+      break;
+    case kLong64_t:
+      branchBuffer = new int64_t;
+      break;
+    case kULong64_t:
+      branchBuffer = new uint64_t;
+      break;
+    case kFloat_t:
+      branchBuffer = new float;
+      break;
+    case kDouble_t:
+      branchBuffer = new double;
+      break;
+    case kBool_t:
+      branchBuffer = new bool;
+      break;
+    default:
+      throw std::runtime_error(
+        std::string{"ROOT_TBranch_ContainerImp::read unsupported fundamental type: "} +
+        DemangleName(type));
+    };
     branchStatus = m_tree->SetBranchAddress(col_name().c_str(),
-                                            reinterpret_cast<void*>(&branchBuffer),
+                                            reinterpret_cast<void*>(branchBuffer),
                                             nullptr,
                                             EDataType(fundInfo->GetType()),
-                                            true);
+                                            false);
   } else {
     auto klass = TClass::GetClass(type);
     if (!klass) {
@@ -74,7 +118,7 @@ bool ROOT_TBranch_Read_ContainerImp::read(int id, void const** data, std::type_i
                                " (col_name='" + col_name() + "', type='" + DemangleName(type) +
                                "')");
     }
-    branchBuffer = nullptr;
+    branchBuffer = klass->New();
     branchStatus = m_tree->SetBranchAddress(
       col_name().c_str(), reinterpret_cast<void*>(&branchBuffer), klass, EDataType::kOther_t, true);
   }
--- a/form/root_storage/root_ttree_write_container.cpp
+++ b/form/root_storage/root_ttree_write_container.cpp
@@ -6,19 +6,13 @@
 #include "TFile.h"
 #include "TTree.h"
 
+#include <gsl/pointers>
+
 using namespace form::detail::experimental;
 
 ROOT_TTree_Write_ContainerImp::ROOT_TTree_Write_ContainerImp(std::string const& name) :
   Storage_Write_Association(name)
 {
-}
-
-ROOT_TTree_Write_ContainerImp::~ROOT_TTree_Write_ContainerImp()
-{
-  if (m_tree != nullptr) {
-    m_tree->GetDirectory()->WriteTObject(m_tree);
-    delete m_tree;
-  }
 }
 
 void ROOT_TTree_Write_ContainerImp::setFile(std::shared_ptr<IStorage_File> file)
@@ -39,10 +33,10 @@ void ROOT_TTree_Write_ContainerImp::setupWrite(std::type_info const& /* type*/)
     throw std::runtime_error("ROOT_TTree_Write_ContainerImp::setupWrite no file attached");
   }
   if (m_tree == nullptr) {
-    m_tree = m_tfile->Get<TTree>(name().c_str());
+    m_tree.reset(m_tfile->Get<TTree>(name().c_str()));
   }
   if (m_tree == nullptr) {
-    m_tree = new TTree(name().c_str(), name().c_str());
+    m_tree.reset(gsl::owner<TTree*>{new TTree(name().c_str(), name().c_str())});
     m_tree->SetDirectory(m_tfile.get());
   }
   if (m_tree == nullptr) {
@@ -61,4 +55,12 @@ void ROOT_TTree_Write_ContainerImp::commit()
   throw std::runtime_error("ROOT_TTree_Write_ContainerImp::commit not implemented");
 }
 
-TTree* ROOT_TTree_Write_ContainerImp::getTTree() { return m_tree; }
+TTree* ROOT_TTree_Write_ContainerImp::getTTree() { return m_tree.get(); }
+
+void ROOT_TTree_Write_ContainerImp::TTreeDeleter::operator()(gsl::owner<TTree*> t) const
+{
+  if (t) {
+    t->GetDirectory()->WriteTObject(t);
+    delete t;
+  }
+}
--- a/form/root_storage/root_ttree_write_container.hpp
+++ b/form/root_storage/root_ttree_write_container.hpp
@@ -16,7 +16,6 @@ namespace form::detail::experimental {
   class ROOT_TTree_Write_ContainerImp : public Storage_Write_Association {
   public:
     ROOT_TTree_Write_ContainerImp(std::string const& name);
-    ~ROOT_TTree_Write_ContainerImp() override;
 
     ROOT_TTree_Write_ContainerImp(ROOT_TTree_Write_ContainerImp const&) = delete;
     ROOT_TTree_Write_ContainerImp& operator=(ROOT_TTree_Write_ContainerImp const&) = delete;
@@ -31,8 +30,16 @@ namespace form::detail::experimental {
     TTree* getTTree();
 
   private:
+    // Be absolutely explicit about the ownership semantics of TTree*,
+    // since ROOT nominally manages the memory of TTree objects, but in
+    // practice we need to ensure objects are written to the TTree, and
+    //  the TTree itself is deleted at the right time.
+    struct TTreeDeleter {
+      void operator()(TTree* t) const;
+    };
+
     std::shared_ptr<TFile> m_tfile;
-    TTree* m_tree{nullptr};
+    std::unique_ptr<TTree, TTreeDeleter> m_tree;
   };
 
 } //namespace form::detail::experimental
--- a/form/root_storage/root_ttree_write_container.hpp
+++ b/form/root_storage/root_ttree_write_container.hpp
@@ -32,12 +32,15 @@ namespace form::detail::experimental {
   private:
     // Be absolutely explicit about the ownership semantics of TTree*,
     // since ROOT nominally manages the memory of TTree objects, but in
-    // practice we need to ensure objects are written to the TTree, and
-    //  the TTree itself is deleted at the right time.
+    // practice we need to ensure objects are written to the TTree—and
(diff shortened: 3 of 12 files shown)


form/root_storage/root_tbranch_read_container.cpp
```diff
@@ -74,7 +74,7 @@ bool ROOT_TBranch_Read_ContainerImp::read(int id, void const** data, std::type_i
                                " (col_name='" + col_name() + "', type='" + DemangleName(type) +
                                "')");
     }
-    branchBuffer = klass->New();
+    branchBuffer = nullptr;
```
pcanal: The mentioned 'risk' of memory is related to the exception thrown on line 83 which indeed should clean up the memory.
This change is a 'disaster' and will lead to the use of delete memory (without additional change)
- yes, TTree will create the allocate the object
- but, TTree will also own it and delete it on line 94
- thus the passing the address up to the caller (line 91) is invalid.

If for some reason this change is wanted, then the `TTree` needs to be informed by setting `branchBuffer` to `nullptr` after line 91.

form/root_storage/root_tbranch_read_container.cpp
```diff
@@ -61,7 +61,7 @@ bool ROOT_TBranch_Read_ContainerImp::read(int id, void const** data, std::type_i
   if (dictInfo->Property() & EProperty::kIsFundamental) {
     //Assume this is a fundamental type like int or double
     auto fundInfo = static_cast<TDataType*>(TDictionary::GetDictionary(type));
-    branchBuffer = new char[fundInfo->Size()];
+    branchBuffer = nullptr;
```
pcanal: This (and the original) code is valid only if the branch is a sub branch (of a top level branch containing an object).
Other wise, it is required to pass the address of the value (and not the address of the pointer)
and ROOT does not take ownership (never deletes it).
If passing the address of a pointer (set to null or otherwise), TTree will write at the address of the pointer itself (erase the memory address value contained in the pointer)
If passing nullptr as the 2nd parameter, ROOT will allocate (and own) the necessary memory but will not (have an indirect way to) return that address.
The allocated address needs to be retrieved by fetching the corresponding `TLeaf` and calling `TLeaf::GetValuePointer` (but the address is owned by the TLeaf and can not be passed along, so it value needs to be retrieve.

Note: currently the interface pass up the address of the data and its ownership, so we need to either contract with Phlex to delete those simple type with `operator delete[] for char` or if Phlex use the correct typed delete operator we need to allocate it appropriately.


So a practical solution is:
```
switch (fundInfo->GetType())
case kDouble_t : branchBuffer = new double;
etc .. See TDataType.h
};

    branchStatus = m_tree->SetBranchAddress(col_name().c_str(),
                                            reinterpret_cast<void*>(branchBuffer),
                                            nullptr,
                                            EDataType(fundInfo->GetType()),
                                            false);
```
The fact that this code is not crashing either means that it is not currently test or that I am misremembering something (or that the test is 'lucky').

form/root_storage/root_tbranch_read_container.cpp
greenc-FNAL: But with `isPtr: true` in the `SetBranchAddress()` call and sending the address of the pointer, wouldn't that have happened anyway? What am I misunderstanding?

test/form/reader.cpp
```diff
@@ -81,11 +82,11 @@ int main(int argc, char** argv)
   for (int nevent = 0; nevent < NUMBER_EVENT; nevent++) {
     std::cout << "PHLEX: Read Event No. " << nevent << std::endl;
 
-    std::vector<float> const* track_x = nullptr;
+    gsl::owner<std::vector<float> const*> track_x = nullptr;
```
pcanal: This will eventually or already clash with the comment on line 199
```
delete track_x; //FIXME: PHLEX owns this memory!
```
and if not, why not use a unique_ptr?

form/root_storage/root_ttree_write_container.hpp
```diff
@@ -29,8 +28,16 @@ namespace form::detail::experimental {
     TTree* getTTree();
 
   private:
+    // Be absolutely explicit about the ownership semantics of TTree*,
+    // since ROOT nominally manages the memory of TTree objects, but in
```
pcanal, line 34: It is an implicit shared ownership.  i.e. `TFile::Close` (called from the `TFile` destructor`) will delete its `TTree` (that are not on the stack) **if** they have not been delete yet.

form/root_storage/root_ttree_write_container.hpp
```diff
@@ -29,8 +28,16 @@ namespace form::detail::experimental {
...
+    // practice we need to ensure objects are written to the TTree, and
+    //  the TTree itself is deleted at the right time.
```
pcanal: This should indicate what the 'right' time is (hint :) : before the corresponding `TFile` is deleted. )

form/root_storage/root_ttree_write_container.hpp
```diff
@@ -29,8 +28,16 @@ namespace form::detail::experimental {
...
+    struct TTreeDeleter {
+      void operator()(TTree* t) const;
+    };
+
     std::shared_ptr<TFile> m_tfile;
-    TTree* m_tree;
+    std::unique_ptr<TTree, TTreeDeleter> m_tree;
   };
```
pcanal, line 46: We may (or may not) want to comment the ordering is critical (`m_tree` must be after `m_tfile` so that it is deleted before `m_tfile`).

form/root_storage/root_tbranch_read_container.cpp
wwuoneway: I will revert this change. As you pointed out, setting `branchBuffer = nullptr `would trigger ROOT's automatic memory management, which could lead to use-after-free or double-free issues if the `TTree` deletes the object while it's still being used upstream. I'll stick to explicit allocation via `klass->New()` to maintain clear ownership.

greenc-FNAL: @wwuoneway As I stated in https://github.com/Framework-R-D/phlex/pull/551#discussion_r3139087677, it is my understanding that with the arguments to `SetBranchAddress()` as they were, the previous situation wasn't correct  either. I would be happy to correct the code, but I'd like to wait until I am clear about what my initial misunderstanding was first.

test/form/reader.cpp
greenc-FNAL: I was assuming that there was a reason for the original use of bare pointer and unadorned new. If that is not the case, then smart pointers would seem to be the way to go.

form/root_storage/root_ttree_write_container.hpp
greenc-FNAL: That is my understanding also, and I was attempting to convey additionally that we wish to control the ordering of the deletion, and ensure that pending objects were written to the tree beforehand. If you have a suggestion for improving the wording, I'd be grateful (notwithstanding the [below](https://github.com/Framework-R-D/phlex/pull/551#discussion_r3139127515)).

form/root_storage/root_tbranch_read_container.cpp
pcanal: Yes, the old version of this line is fine. (Changing it to the proposed version is possible but the pointer needs to be set to nullptr at the end).

plugins/python/src/modulewrap.cpp
```diff
@@ -943,15 +944,16 @@ static PyObject* md_transform(py_phlex_module* mod, PyObject* args, PyObject* kw
 
   switch (input_queries.size()) {
   case 1: {
-    auto* pyc = new py_callback_1{callable}; // TODO: leaks, but has program lifetime
+    gsl::owner<py_callback_1*> pyc =
+      new py_callback_1{callable}; // TODO: leaks, but has program lifetime
     mod->ph_module->transform(pyname, *pyc, concurrency::serial)
       .input_family(
```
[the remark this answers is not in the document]
greenc-FNAL: Per the `TODO`, these program-lifetime leaks are intentional at this time.

plugins/python/src/modulewrap.cpp
```diff
@@ -1032,14 +1034,14 @@ static PyObject* md_observe(py_phlex_module* mod, PyObject* args, PyObject* kwds
 
   switch (input_queries.size()) {
   case 1: {
-    auto* pyc = new py_callback_1v{callable};
+    gsl::owner<py_callback_1v*> pyc = new py_callback_1v{callable};
     mod->ph_module->observe(cname, *pyc, concurrency::serial)
       .input_family(
         product_query{.creator = identifier(c0), .layer = pq0.layer, .suffix = identifier(suff0)});
```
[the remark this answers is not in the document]
greenc-FNAL: Per the `TODO`, these program-lifetime leaks are intentional at this time.

plugins/python/src/modulewrap.cpp
```diff
@@ -1237,25 +1239,25 @@ static PyObject* sc_provide(py_phlex_source* src, PyObject* args, PyObject* kwds
   // is fixed, so there is no combinatorics problem.
   std::string const& out_type = output_types[0];
   if (out_type == "bool") {
-    auto* pyc = new provider_cb_bool{callable};
+    gsl::owner<provider_cb_bool*> pyc = new provider_cb_bool{callable};
     src->ph_source->provide(functor_name, *pyc).output_product(opq.value());
   } else if (out_type == "int32_t") {
```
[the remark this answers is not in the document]
greenc-FNAL: Per the `TODO`, these program-lifetime leaks are intentional at this time.

form/root_storage/root_tbranch_read_container.cpp
```diff
@@ -61,12 +61,56 @@ bool ROOT_TBranch_Read_ContainerImp::read(int id, void const** data, std::type_i
...
+      branchBuffer = new unsigned int;
+      break;
+    case kLong_t:
+      branchBuffer = new long;
+      break;
+    case kULong_t:
+      branchBuffer = new unsigned long;
+      break;
+    case kLong64_t:
+      branchBuffer = new int64_t;
```
pcanal:
```suggestion
      branchBuffer = new Long64_t;
```
to fully match the ROOT type (on some platform `int64_t` is `long`) on all platform `Long64_t` is `long long`.

form/root_storage/root_tbranch_read_container.cpp
```diff
@@ -61,12 +61,56 @@ bool ROOT_TBranch_Read_ContainerImp::read(int id, void const** data, std::type_i
...
+      branchBuffer = new bool;
+      break;
+    default:
+      throw std::runtime_error(
+        std::string{"ROOT_TBranch_ContainerImp::read unsupported fundamental type: "} +
+        DemangleName(type));
+    };
     branchStatus = m_tree->SetBranchAddress(col_name().c_str(),
-                                            reinterpret_cast<void*>(&branchBuffer),
+                                            reinterpret_cast<void*>(branchBuffer),
```
pcanal: I am still a bit curious why the previous code is not causing problem in testing.

form/root_storage/root_ttree_write_container.cpp
```diff
@@ -39,10 +33,10 @@ void ROOT_TTree_Write_ContainerImp::setupWrite(std::type_info const& /* type*/)
     throw std::runtime_error("ROOT_TTree_Write_ContainerImp::setupWrite no file attached");
   }
   if (m_tree == nullptr) {
-    m_tree = m_tfile->Get<TTree>(name().c_str());
+    m_tree.reset(m_tfile->Get<TTree>(name().c_str()));
   }
   if (m_tree == nullptr) {
-    m_tree = new TTree(name().c_str(), name().c_str());
+    m_tree.reset(gsl::owner<TTree*>{new TTree(name().c_str(), name().c_str())});
```
knoepfel, line 39:
```suggestion
    m_tree = std::make_unique<TTree>(name().c_str(), name().c_str());
```

greenc-FNAL: @knoepfel There appears to be no viable replacement assignment, so I backed out your suggested change.

```console
[7/34] Building CXX object form/root_storage/CMakeFiles/root_storage.dir/root_ttree_write_container.cpp.o
FAILED: form/root_storage/CMakeFiles/root_storage.dir/root_ttree_write_container.cpp.o 
/opt/spack-environments/phlex-ci/.spack-env/view/bin/clang++ -DUSE_ROOT_STORAGE -I/workspaces/phlex/form -isystem /opt/spack-environments/phlex-ci/.spack-env/view/include/root -isystem /workspaces/phlex/build-clang-tidy/_deps/gsl-src/include -Og -g -std=c++23 -fPIC -Wall -Werror -Wunused -Wunused-parameter -pedantic -MD -MT form/root_storage/CMakeFiles/root_storage.dir/root_ttree_write_container.cpp.o -MF form/root_storage/CMakeFiles/root_storage.dir/root_ttree_write_container.cpp.o.d -o form/root_storage/CMakeFiles/root_storage.dir/root_ttree_write_container.cpp.o -c /workspaces/phlex/form/root_storage/root_ttree_write_container.cpp
/workspaces/phlex/form/root_storage/root_ttree_write_container.cpp:39:12: error: no viable overloaded '='
   39 |     m_tree = std::make_unique<TTree>(name().c_str(), name().c_str());
      |     ~~~~~~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/spack/opt/spack/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placehold/linux-x86_64_v3/gcc-15.2.0-aypjjrmq235e2xbp2siw3uqu76kk442t/lib/gcc/x86_64-pc-linux-gnu/15.2.0/../../../../include/c++/15.2.0/bits/unique_ptr.h:409:19: note: candidate function not viable: no known conversion from 'unique_ptr<[...], (default) std::default_delete<TTree>>' to 'unique_ptr<[...], form::detail::experimental::ROOT_TTree_Write_ContainerImp::TTreeDeleter>' for 1st argument
  409 |       unique_ptr& operator=(unique_ptr&&) = default;
      |                   ^         ~~~~~~~~~~~~
/spack/opt/spack/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placehold/linux-x86_64_v3/gcc-15.2.0-aypjjrmq235e2xbp2siw3uqu76kk442t/lib/gcc/x86_64-pc-linux-gnu/15.2.0/../../../../include/c++/15.2.0/bits/unique_ptr.h:435:7: note: candidate function not viable: no known conversion from '__detail::__unique_ptr_t<TTree>' (aka 'unique_ptr<TTree>') to 'nullptr_t' (aka 'std::nullptr_t') for 1st argument
  435 |       operator=(nullptr_t) noexcept
      |       ^         ~~~~~~~~~
/spack/opt/spack/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placehold/linux-x86_64_v3/gcc-15.2.0-aypjjrmq235e2xbp2siw3uqu76kk442t/lib/gcc/x86_64-pc-linux-gnu/15.2.0/../../../../include/c++/15.2.0/bits/unique_ptr.h:526:19: note: candidate function not viable: no known conversion from 'unique_ptr<[...], (default) std::default_delete<TTree>>' to 'const unique_ptr<[...], form::detail::experimental::ROOT_TTree_Write_ContainerImp::TTreeDeleter>' for 1st argument
  526 |       unique_ptr& operator=(const unique_ptr&) = delete;
      |                   ^         ~~~~~~~~~~~~~~~~~
/spack/opt/spack/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placehold/linux-x86_64_v3/gcc-15.2.0-aypjjrmq235e2xbp2siw3uqu76kk442t/lib/gcc/x86_64-pc-linux-gnu/15.2.0/../../../../include/c++/15.2.0/bits/unique_ptr.h:425:2: note: candidate template ignored: requirement 'std::__and_<std::__and_<std::is_convertible<TTree *, TTree *>, std::__not_<std::is_array<TTree>>>, std::is_assignable<form::detail::experimental::ROOT_TTree_Write_ContainerImp::TTreeDeleter &, std::default_delete<TTree> &&>>::value' was not satisfied [with _Up = TTree, _Ep = std::default_delete<TTree>]
  425 |         operator=(unique_ptr<_Up, _Ep>&& __u) noexcept
      |         ^
1 error generated.
[22/34] Building CXX object plugins/python/CMakeFiles/pymodule.dir/src/modulewrap.cpp.o
ninja: build stopped: subcommand failed.
```

test/form/form_storage_test.cpp
```diff
@@ -111,3 +97,247 @@ TEST_CASE("FORM Container setup error handling")
...
+namespace {
+  char const* const kFundTestTree = "FundTestTree";
+
+  template <typename T>
+  void writeFundamentalDirect(std::string const& fileName,
+                              std::string const& branchName,
+                              std::string const& leafSpec,
+                              T value)
+  {
+    TFile f(fileName.c_str(), "RECREATE");
```
pcanal, line 119: Shouldn't this code use FORM to write the file or is the intent to check that FORM can read this specific format?

form/root_storage/root_tbranch_read_container.cpp
pcanal: It was indeed NOT tested properly.  It was executed only in one test that was checking that it threw an exception in case of incorrect outer setup but was never testing that when properly setup it was doing the right things.  This fixed by https://github.com/Framework-R-D/phlex/pull/551/changes/3ccbc8c12954e1d4758abf4e38f1b6ac7242c528

form/root_storage/root_tbranch_read_container.cpp
```diff
@@ -61,20 +63,61 @@ bool ROOT_TBranch_Read_ContainerImp::read(int id, void const** data, std::type_i
   if (dictInfo->Property() & EProperty::kIsFundamental) {
     //Assume this is a fundamental type like int or double
     auto fundInfo = static_cast<TDataType*>(TDictionary::GetDictionary(type));
-    branchBuffer = new char[fundInfo->Size()];
-    branchStatus = m_tree->SetBranchAddress(col_name().c_str(),
-                                            reinterpret_cast<void*>(&branchBuffer),
-                                            nullptr,
-                                            EDataType(fundInfo->GetType()),
-                                            true);
+    switch (fundInfo->GetType()) {
```
aolivier23, line 66: I like the explicit types here instead of an array of `char`.  But I'm concerned that this is harder to maintain than the old version.  What are we gaining here to offset the maintenance cost?

form/root_storage/root_tbranch_read_container.cpp
```diff
@@ -61,20 +63,61 @@ bool ROOT_TBranch_Read_ContainerImp::read(int id, void const** data, std::type_i
...
+      col_name().c_str(), branchBuffer, nullptr, EDataType(fundInfo->GetType()), false);
   } else {
     auto klass = TClass::GetClass(type);
     if (!klass) {
       throw std::runtime_error(std::string{"ROOT_TBranch_ContainerImp::read missing TClass"} +
                                " (col_name='" + col_name() + "', type='" + DemangleName(type) +
                                "')");
     }
-    branchBuffer = klass->New();
+    branchBuffer = gsl::owner<void*>(klass->New());
```
wwuoneway, line 120: The explicit `gal::owner<void*>` cast is redundant since `branchBuffer` is already declared as on [Line 54](https://github.com/Framework-R-D/phlex/pull/551/changes#diff-d713cf1ad10ac0fb5c742c7ea5a431fefa3481cd4538c55932b3a9ee1c0c881aR54). Consider simplifying to:

`branchBuffer = klass->New();`

form/root_storage/root_tbranch_read_container.cpp
pcanal: The caller owns the memory and needs to delete it.   The question is then whether the callers should do the 'obvious':
```
delete (real_type*)ptr;
```
or
```
delete [] (char*)ptr;
```
where the former is *necessary* for structured type and thus the caller would have to treat them differently.

test/form/form_storage_test.cpp
aolivier23: I agree.  Is there a way to rewrite `test_utils.hpp` so we only have one function that works for fundamental types too?  I'll consider that homework for me too.

form/root_storage/root_tbranch_read_container.cpp
aolivier23: I didn't realize that `delete (int*)ptr;` won't work correctly if I allocate space for the `int` as a `char` array of `sizeof(int)`.  I definitely don't intend for the caller to have to delete as an array.

form/root_storage/root_tbranch_read_container.cpp
greenc-FNAL: The explicit cast is required by `clang-tidy`:

```console
[1/1][1.2s] /opt/spack-environments/phlex-ci/.spack-env/view/bin/clang-tidy -checks=-*,cppcoreguidelines-owning-memory -export-fixes /tmp/tmphtrqidc0/tmpgj1whkmt.yaml -p=/workspaces/phlex/build-clang-tidy /workspaces/phlex/form/root_storage/root_tbranch_read_container.cpp
/workspaces/phlex/form/root_storage/root_tbranch_read_container.cpp:120:5: warning: expected assignment source to be of type 'gsl::owner<>'; got 'void *' [cppcoreguidelines-owning-memory]
  120 |     branchBuffer = klass->New();
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~
1 warning generated.
```

test/form/form_storage_test.cpp
greenc-FNAL: @aolivier23 @pcanal Is that a task for this PR, or afterwards?

form/root_storage/root_tbranch_read_container.cpp
greenc-FNAL: @aolivier23 @pcanal Is what is in the PR currently correct and reasonable code, or does something still need to be tweaked here?

aolivier23: I never made any progress with researching why `delete (int*)ptr;` from the example above would be incorrect.  Is it correct that `delete [] (char*)ptr;` is necessary with the old way this code worked?  If so, then I support this change and am willing to mark as resolved.

test/form/form_storage_test.cpp
aolivier23: @greenc-FNAL Thank you for revisiting this.  What's the reason we prefer not to use FORM to do this write?  I worry this PR makes this test fragile against FORM changing how it names branches.

pcanal: @aolivier23 Note that this is one of the test from FORM itself, so indeed this would need to be changed once the format produced by FORM changes.  We can address this in X ways:
(a) call FORM here instead of hand coded, do this change later
(b) same but do this right now before this PR is merged.
(c) leave as is until there is a change in layout and then update as needed (eg calling FORM here too).

Which of these three options would you prefer (and whom do you prefer does the update)?

form/root_storage/root_tbranch_read_container.cpp
pcanal: See https://isocpp.org/wiki/faq/freestore-mgmt#delete-array which is pretty (probably too much) emphatic about it :).

For a practical example, implementer are completely allowed to have the allocation done by `operator new` and by `operator new []` to come from two completely distinct memory 'stash' and then do the 'right' thing when the corresponding `operator delete` or `delete []` is called (for example putting it back on the free memory list of that `stash` and coalesce).

`operator new []` also sometimes/often stash the allocated size somewhere and this need to be freed too.  For example it might be put 'in front' of the memory address returned to the user and thus the actual address to be 'freed' is lower than the one given by the user to operator delete.

pcanal (dismissed):
Looks good to me.  If https://github.com/Framework-R-D/phlex/pull/551/changes#r3157633667 is not resolved in this PR, there ought to be an issue open to keep track of its resolution.

test/form/form_storage_test.cpp
wwuoneway: @pcanal thanks. I would prefer addressing this later, so option (a) sounds good to me. We will provide the update on this in a separate task later.

wwuoneway (dismissed):
LGTM. https://github.com/Framework-R-D/phlex/pull/551/changes#r3157633667 would need a separate issue track.

test/form/form_storage_test.cpp
greenc-FNAL: The test here is specifically testing the correct behavior of code in `root_storage/root_tbranch_read_container.cpp::read()`. Is the issue that the test is in the wrong place, or that there should be a "generic FORM" test to complement it? I'm sorry I'm not getting the thrust of your discussion, here.

aolivier23: I agree with option a).  I'll put some work into updating after this PR.  Thank you for discussing.

form/root_storage/root_tbranch_read_container.cpp
aolivier23: I agree to let this change go through.  Thank you for explaining and the good reference :)

knoepfel (approved):

PR #552, opened 2026-04-26
Revised implementation of FORM source infrastructure

wwuoneway:
**Overview**
This PR introduces a revised implementation of the FORM Source infrastructure. It replaces #379 to resolve stale code conflicts and adopt a cleaner, class-based architecture that aligns with current Phlex standards.

**Key Changes**
- **Infrastructure Refactoring**: Introduced FormInputSource to manage the reader lifecycle, ensuring robust resource handling.
- **Dynamic Registration**: Implemented s.provide logic within PHLEX_REGISTER_PROVIDERS to bridge FORM data products into the Phlex graph.
- **Clean Slate**: This PR is built on a fresh branch to ensure zero conflicts with deprecated logic from previous iterations.

**Codecov Note**
The **Patch Coverage** is currently 0.0% as this PR focuses on the internal infrastructure for FormInputSource. Full integration tests will be provided in a follow-up PR once data retrieval backend and configuration testing are fully implemented and validated.

### DIFF
--- a/form/form_source.cpp
+++ b/form/form_source.cpp
@@ -24,7 +24,7 @@ namespace {
     }
 
     template <typename T>
-    T read(std::string const& creator,
+    T const& read(std::string const& creator,
            std::string const& product_name,
            phlex::data_cell_index const& id)
     {
--- a/form/form_source.cpp
+++ b/form/form_source.cpp
@@ -1,0 +1,98 @@
+#include "phlex/source.hpp"
+
+#include "form/config.hpp"
+#include "form/form_reader.hpp"
+#include "form/technology.hpp"
+
+#include <cassert>
+#include <iostream>
+#include <memory>
+#include <stdexcept>
+#include <string>
+#include <string_view>
+#include <unordered_map>
+#include <vector>
+
+namespace {
+
+  class FormInputSource {
+  public:
+    FormInputSource(form::experimental::config::ItemConfig const& input_cfg,
+                    form::experimental::config::tech_setting_config const& tech_cfg) :
+      reader_(std::make_shared<form::experimental::form_reader_interface>(input_cfg, tech_cfg))
+    {
+    }
+
+    template <typename T>
+    T read(std::string const& creator,
+           std::string const& product_name,
+           phlex::data_cell_index const& id)
+    {
+      std::string const index_str = id.to_string();
+      form::experimental::product_with_name pb{product_name, nullptr, &typeid(T)};
+      reader_->read(creator, index_str, pb);
+      if (!pb.data) {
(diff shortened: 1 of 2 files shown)


form/form_source.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+}
+
+PHLEX_REGISTER_PROVIDERS(s, config)
+{
+  std::cout << "Registering FORM input source...\n";
+
+  // Extract configuration from Phlex config
+  std::string const input_file = config.get<std::string>("input_file");
+  std::string const creator = config.get<std::string>("creator");
```
gemmeren, line 53: Just a note that I don't think its good to require creator, technology to be configured, those should come from the input file.

form/form_source.cpp
```diff
@@ -0,0 +1,98 @@
...
+  auto form_input = std::make_shared<FormInputSource>(input_cfg, tech_cfg);
+
+  // --- Register providers dynamically from config ---
+  // FIXME: Prototype 0.1 -- types hardcoded as int.
+  for (auto const& name : products) {
+    s.provide("provide_" + name,
+              [form_input, creator, name](phlex::data_cell_index const& id) -> int {
+                return form_input->read<int>(creator, name, id);
+              })
+      .output_product(phlex::product_query{.creator = phlex::experimental::identifier(creator),
```
gemmeren, line 92: Why is source dealing with output_products?

form/form_source.cpp
wwuoneway: thanks, I agree and plan to implement this feature in the coming PR.

form/form_source.cpp
knoepfel: `output_product(...)` here just refers to the output of the provider node, not output from the job.

wwuoneway: The naming may seem counterintuitive since we are implementing the reading feature through a source, but `.output_product` is a requirement of the graph provider pattern used by the plex framework. Here's my understanding: `form_source` performs **I/O input** from the perspective of the application, it acts as a **producer** from the perspective of the graph. The `.output_product` call is the mechanism for a provider to "advertise" its capabilities. It declares the **output schema**, so the framework can build the dependency map. So in short, the term "output_product" refers to the node's role in the data flow rather than the direction of the disk I/O.

aolivier23 (dismissed):
Looks good to me.  Please make sure we check on memory usage in the follow-up testing PR.

aolivier23 (approved):
Looks good to me.  Please make sure we double-check memory usage in the follow-up testing PR.  Thank you for getting this across the finish line.

PR #553, opened 2026-04-28
`clang-tidy`: resolve `cppcoreguidelines-pro-bounds-avoid-unchecked-container-access`

greenc-FNAL:
- **Use `front()` for first element access**
- **Use ranges/range-for to replace unchecked access**
- **Use structured bindings/`generate_n()` to replace unchecked access**
- **Ignore associative element access**
- **Use `at()` to replace unchecked access**

### DIFF
--- a/phlex/core/detail/filter_impl.cpp
+++ b/phlex/core/detail/filter_impl.cpp
@@ -79,12 +79,11 @@ namespace phlex::experimental {
     }
 
     // Fill slots in the order of the input arguments to the downstream node.
-    for (auto ip_it = input_products_->cbegin(); auto& e : elem) {
-      auto const& ip = *ip_it++;
-      if (e or not resolve_in_store(ip, *store)) {
+    for (std::size_t i = 0; i != nargs_; ++i) {
+      if (elem[i] or not resolve_in_store((*input_products_)[i], *store)) {
         continue;
       }
-      e = store;
+      elem[i] = store;
     }
   }
 
--- a/phlex/core/detail/filter_impl.cpp
+++ b/phlex/core/detail/filter_impl.cpp
@@ -79,11 +79,12 @@ namespace phlex::experimental {
     }
 
     // Fill slots in the order of the input arguments to the downstream node.
-    for (std::size_t i = 0; i != nargs_; ++i) {
-      if (elem[i] or not resolve_in_store((*input_products_)[i], *store)) {
+    for (auto ip_it = input_products_->cbegin(); auto& e : elem) {
+      auto const& ip = *ip_it++;
+      if (e or not resolve_in_store(ip, *store)) {
         continue;
       }
-      elem[i] = store;
+      e = store;
     }
   }
 
--- a/phlex/model/product_matcher.cpp
+++ b/phlex/model/product_matcher.cpp
@@ -45,12 +45,13 @@ namespace phlex::experimental {
   {
   }
 
-  product_matcher::product_matcher(std::array<std::string, 4u> fields) :
-    layer_path_{std::move(fields[0])},
-    module_name_{std::move(fields[1])},
-    node_name_{std::move(fields[2])},
-    product_name_{std::move(fields[3])}
+  product_matcher::product_matcher(std::array<std::string, 4u> fields)
   {
+    auto& [lp, mn, nn, pn] = fields;
+    layer_path_ = std::move(lp);
+    module_name_ = std::move(mn);
+    node_name_ = std::move(nn);
+    product_name_ = std::move(pn);
   }
 
   std::string product_matcher::encode() const
--- a/plugins/layer_generator.cpp
+++ b/plugins/layer_generator.cpp
@@ -76,6 +76,7 @@ namespace phlex::experimental {
   {
     // First check if layer paths need to be rebased
     std::vector<size_t> indices_for_rebasing;
+    // We can use std::views::enumerate once the AppleClang C++ STL supports it.
     for (std::size_t i = 0ull, n = layer_paths_.size(); i != n; ++i) {
       auto const& layer = layer_paths_[i];
       if (layer.starts_with("/" + layer_name)) {
--- a/plugins/python/src/configwrap.cpp
+++ b/plugins/python/src/configwrap.cpp
@@ -1,5 +1,6 @@
 #include <cstdint>
 #include <optional>
+#include <ranges>
 #include <string>
 
 #include "phlex/configuration.hpp"
--- a/plugins/python/src/configwrap.cpp
+++ b/plugins/python/src/configwrap.cpp
@@ -94,6 +94,7 @@ static PyObject* pcm_subscript(py_config_map* pycmap, PyObject* pykey)
         if (!cvalue_size)
           return nullptr;
         pyvalue = PyTuple_New(*cvalue_size);
+        // We can use std::views::enumerate once the AppleClang C++ STL supports it.
         for (Py_ssize_t i = 0; i < *cvalue_size; ++i) {
           PyObject* item = PyLong_FromLong((long)cvalue[i]);
           PyTuple_SetItem(pyvalue, i, item);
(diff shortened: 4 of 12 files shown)


test/repeater_node_test.cpp
```diff
@@ -273,8 +273,8 @@ TEST_CASE("Test repeater data before index", "[multithreading]")
   // The messages are not guaranteed to arrive at a particular order, so we sort by message ID
   // before checking.
   CHECK(fixture.sorted_message_ids() == std::vector<std::size_t>{1, 2});
-  CHECK(fixture.consumed_messages()[0].store == store1);
-  CHECK(fixture.consumed_messages()[1].store == store1);
+  CHECK(fixture.consumed_messages().at(0).store == store1);
+  CHECK(fixture.consumed_messages().at(1).store == store1);
```
knoepfel: Size already checked in line 272

test/repeater_node_test.cpp
```diff
@@ -246,7 +246,7 @@ TEST_CASE("Test repeater transition to pass-through", "[multithreading]")
   fixture.wait_for_all();
 
   CHECK(fixture.consumed_messages().size() == 2);
-  CHECK(fixture.consumed_messages()[1].store == store2);
+  CHECK(fixture.consumed_messages().at(1).store == store2);
```
knoepfel: Size already checked in line 248.  Suppose we could change line 248 to `REQUIRE(...)`

test/repeater_node_test.cpp
```diff
@@ -236,7 +236,7 @@ TEST_CASE("Test repeater transition to pass-through", "[multithreading]")
   fixture.wait_for_all();
 
   CHECK(fixture.consumed_messages().size() == 1);
-  CHECK(fixture.consumed_messages()[0].store == store1);
+  CHECK(fixture.consumed_messages().at(0).store == store1);
```
knoepfel: Size already checked in line 238 (could replace that check with `REQUIRE(...)`

test/repeater_node_test.cpp
knoepfel, line 1: All direct accesses here are correct.  We can consider replacing the `CHECK`s on sizes to `REQUIRE`s to make that more explicit.

test/core_misc_test.cpp
```diff
@@ -104,5 +106,5 @@ TEST_CASE("add_to_error_messages tests", "[core]")
   add_to_error_messages(errors, "duplicate_node");
 
   REQUIRE(errors.size() == 1);
-  CHECK(errors[0].find("duplicate_node") != std::string::npos);
+  CHECK(errors.at(0).find("duplicate_node") != std::string::npos);
```
knoepfel: Already guaranteed that the size is correct. based on `REQUIRE(errors.size() == 1)`.

test/form/toy_tracker.cpp
```diff
@@ -1,20 +1,21 @@
...
-                 static_cast<float>(generateRandom()) / static_cast<float>(random_max),
-                 static_cast<float>(generateRandom()) / static_cast<float>(random_max));
-  }
+  std::vector<TrackStart> points;
+  points.reserve(npx);
+  std::generate_n(std::back_inserter(points), npx, [this] {
+    return TrackStart(static_cast<float>(generateRandom()) / static_cast<float>(random_max),
+                      static_cast<float>(generateRandom()) / static_cast<float>(random_max),
+                      static_cast<float>(generateRandom()) / static_cast<float>(random_max));
+  });
```
knoepfel, line 18: Unnecessary change–`nelement` will never exceed `npx` here.

test/form/form_basics_test.cpp
knoepfel, line 1: All direct accesses here are correct and already guarded by an appropriate `REQUIRE` macro.

form/form/form_writer.cpp
knoepfel, line 1: Direct access is correct here and will not fail due to the check on line 50

phlex/core/detail/filter_impl.cpp
knoepfel, line 1: Direct access in this file is correct and is guaranteed to not fail due to line 70

phlex/core/index_router.cpp
```diff
@@ -269,7 +269,7 @@ namespace phlex::experimental {
     }
 
     if (candidates.size() == 1ull) {
-      return candidates[0]->second;
+      return candidates.front()->second;
```
knoepfel: Direct access is guaranteed to be safe based on the check in line 271

phlex/core/input_arguments.hpp
```diff
@@ -45,14 +45,16 @@ namespace phlex::experimental {
           query,
           fmt::join(products | views::transform(&product_specification::full), "\n    - ")));
       }
-      return store->get_handle<handle_arg_t>(products[0]);
+      return store->get_handle<handle_arg_t>(products.front());
```
knoepfel: Direct access is safe here due to the check on line 34

phlex/model/data_layer_hierarchy.cpp
```diff
@@ -64,7 +64,7 @@ namespace phlex::experimental {
       throw std::runtime_error(msg);
     }
 
-    return candidates[0]->count.load();
+    return candidates.front()->count.load();
```
knoepfel: Direct access is safe here due to the check on line 53

phlex/model/fixed_hierarchy.cpp
```diff
@@ -20,7 +20,7 @@ namespace {
     if (path.empty()) {
       throw std::runtime_error("Layer paths cannot be empty.");
     }
-    auto const rest = std::span{path}.subspan(path[0] == "job" ? 1 : 0);
+    auto const rest = std::span{path}.subspan(path.front() == "job" ? 1 : 0);
```
knoepfel: Direct access is safe here due to the check on line 20

phlex/model/product_matcher.cpp
```diff
@@ -16,7 +16,7 @@ namespace {
     if (empty(matcher_spec)) {
       throw std::runtime_error("Empty product-matcher specifications are not allowed.");
     }
-    if (matcher_spec[0] == '/') {
+    if (matcher_spec.front() == '/') {
```
knoepfel: Direct access is safe here due to the check on line 16

phlex/model/products.hpp
```diff
@@ -57,15 +57,17 @@ namespace phlex::experimental {
     void add_all(product_specifications const& names, Ts ts)
     {
       assert(names.size() == 1ull);
-      add(names[0], std::move(ts));
+      add(names.front(), std::move(ts));
```
knoepfel: Direct access is assumed to be safe here due to the precondition on line 59

plugins/python/src/configwrap.cpp
knoepfel, line 1: `std::views::enumerate` is not yet supported with `apple-clang`.  Please revert

plugins/layer_generator.cpp
```diff
@@ -76,17 +76,16 @@ namespace phlex::experimental {
   {
     // First check if layer paths need to be rebased
     std::vector<size_t> indices_for_rebasing;
-    for (std::size_t i = 0ull, n = layer_paths_.size(); i != n; ++i) {
-      auto const& layer = layer_paths_[i];
+    for (auto const [i, layer] : layer_paths_ | std::views::enumerate) {
```
knoepfel: Enumerate not yet supported on `apple-clang`, please revert.

knoepfel (commented):
Unfortunately, I think we may need to disable the `cppcoreguidelines-pro-bounds-avoid-unchecked-container-access`.  There might be _a few_ places in this PR where switching to `.front()` or `.at(...)` is a benefit.  But the vast majority of cases this PR changes are places where direct use of unchecked access is guaranteed (or assumed, based on preconditiions) to be successful based on previous checks in the code.  I note this in various comments below.  For efficiency reasons, I am very loath to abandon direct access there.

It is possible to annotate uses of `operator[]` (as has been done for some cases), but doing that everywhere where direct access is efficient seems like overkill.  I think this is a case where we rely on our unit tests (coupled with the code coverage checks currently in place) to make sure that the behavior is correct.

I will note that switching to `std::views::zip` is welcome.  But I would not switch to `std::views::enumerate` until `apple-clang` supports it (Apple clang 21.0.0 does not).

test/form/toy_tracker.cpp
greenc-FNAL: True, but here you're only reserving the space rather than default-constructing `npx` elements and then constructing `npx` more and copy-assigning them.

knoepfel: That's fair.  No need to change it back.

knoepfel (approved):

PR #554, opened 2026-04-28
`clang-tidy`: resolve `cppguidelines-pro-type-*cast*`

greenc-FNAL:
- **Address review comment**
- **Ignore some `clang-tidy` checks by directory**
- **Annotate known/verified instance of `static_cast` to derived**

### DIFF
--- a/plugins/python/src/configwrap.cpp
+++ b/plugins/python/src/configwrap.cpp
@@ -20,16 +20,16 @@ struct phlex::experimental::py_config_map {
 PyObject* phlex::experimental::wrap_configuration(configuration const& config)
 {
   py_config_map* pyconfig =
-    (py_config_map*)PhlexConfig_Type.tp_new(&PhlexConfig_Type, nullptr, nullptr);
+    reinterpret_cast<py_config_map*>(PhlexConfig_Type.tp_new(&PhlexConfig_Type, nullptr, nullptr));
 
   pyconfig->ph_config = &config;
 
-  return (PyObject*)pyconfig;
+  return reinterpret_cast<PyObject*>(pyconfig);
 }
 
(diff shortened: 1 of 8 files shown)


plugins/python/src/configwrap.cpp
```diff
@@ -213,17 +213,18 @@ static PyObject* pcm_subscript(py_config_map* pycmap, PyObject* pykey)
...
+static PyMappingMethods pcm_as_mapping = {
+  nullptr, reinterpret_cast<binaryfunc>(pcm_subscript), nullptr};
 
 // clang-format off
 // PyType_Ready() modifies PyTypeObject in-place; the Python C API requires non-const.
 // NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
 PyTypeObject phlex::experimental::PhlexConfig_Type = {
   PyVarObject_HEAD_INIT(&PyType_Type, 0)
-  (char*) "pyphlex.configuration",                   // tp_name
+  const_cast<char*>("pyphlex.configuration"),        // tp_name
```
knoepfel: If I remember correctly, I _think_ that each of these hard-casts (now `const_cast`s) can simply be replaced by the string literal:

```suggestion
  "pyphlex.configuration",                           // tp_name
```

Can we try making that change here and elsewhere?  If that works, we'd be able to re-enable the `cppcoreguidelines-pro-type-const-cast` check for this directory.

knoepfel: Can we do a more complete survey of uses of `const_cast<char*>("x")`, and just replace them with `"x"`?  I did not intend my comment above to apply to only one place:

> Can we try making that change here and elsewhere? If that works, we'd be able to re-enable the `cppcoreguidelines-pro-type-const-cast` check for this directory.

knoepfel (approved):

PR #572, opened 2026-05-07
Replace `char const*` with `char*` as appropriate

knoepfel:
This PR fixes a compilation failure when permissive conversions are not enabled.

### DIFF
--- a/plugins/python/src/modulewrap.cpp
+++ b/plugins/python/src/modulewrap.cpp
@@ -679,7 +679,8 @@ static PyObject* parse_args(PyObject* args,
 
   static char kw0[] = "callable", kw1[] = "input_family", kw2[] = "output_product_suffixes",
               kw3[] = "concurrency", kw4[] = "name";
-  static char const* kwnames[] = {kw0, kw1, kw2, kw3, kw4, nullptr};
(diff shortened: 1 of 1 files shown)


greenc-FNAL:
Is there something we need to do to CI and/or the CMake defaults to ensure that this is caught in the future before the offending PR is merged?

knoepfel:
> Is there something we need to do to CI and/or the CMake defaults to ensure that this is caught in the future before the offending PR is merged?

Claude Sonnet 4.6 points at [the following change in the Python C API ](https://docs.python.org/3.13/whatsnew/3.13.html#changed-c-apis):

> - Python 3.12 declares `PyArg_ParseTupleAndKeywords` (in C++) as `(..., char**, ...)`. A `const char**` argument cannot implicitly convert to `char**` — this is a hard error.
> - Python 3.13 declares it (in C++) as `(..., const char *const*, ...)`. A `const char**` (pointer to pointer to `const char`) can implicitly convert to `const char *const*` (pointer to `const` pointer to `const char`) — adding `const` at the outer pointer level is a valid qualification conversion per C++11 §4.4 [conv.qual].

pcanal (approved):
LGTM.

greenc-FNAL (approved):

PR #582, opened 2026-05-13
Optimize FORM storage init by replacing double find with insert(...).first

wwuoneway:
## Summary

Replace double `find()` lookups with single-lookup using `insert(...).first` in FORM storage initialization.

**Changes**
- Before: `find()` → `insert()` → `find()` again (two lookups)
- After: `find()` → `insert(...).first` (single lookup)

**Why not `try_emplace`?**
An earlier version attempted `try_emplace` but caused test failures due to unconditional constructor calls that broke lazy initialization semantics. The `insert(...).first` pattern is safe and equally efficient.

**Files changed:** `storage_writer.cpp`, `storage_reader.cpp`

### DIFF
--- a/form/storage/storage_reader.cpp
+++ b/form/storage/storage_reader.cpp
@@ -24,21 +24,21 @@ int StorageReader::getIndex(Token const& token,
     auto key = std::make_pair(token.fileName(), token.containerName());
     auto cont = m_read_containers.find(key);
     if (cont == m_read_containers.end()) {
-      auto [file, inserted] = m_files.try_emplace(
-        token.fileName(), createFile(token.technology(), token.fileName(), 'i'));
-      if (inserted) {
+      auto file = m_files.find(token.fileName());
+      if (file == m_files.end()) {
+        file =
+          m_files.insert({token.fileName(), createFile(token.technology(), token.fileName(), 'i')})
+            .first;
         for (auto const& [key, value] : settings.getFileTable(token.technology(), token.fileName()))
           file->second->setAttribute(key, value);
       }
-      auto [cont_it, cont_inserted] = m_read_containers.try_emplace(
-        key, createReadContainer(token.technology(), token.containerName()));
-      if (cont_inserted) {
-        for (auto const& [key, value] :
-             settings.getContainerTable(token.technology(), token.containerName()))
-          cont_it->second->setAttribute(key, value);
-        cont_it->second->setFile(file->second);
-      }
-      cont = cont_it;
+      cont = m_read_containers
+               .insert({key, createReadContainer(token.technology(), token.containerName())})
+               .first;
+      for (auto const& [key, value] :
+           settings.getContainerTable(token.technology(), token.containerName()))
+        cont->second->setAttribute(key, value);
+      cont->second->setFile(file->second);
     }
     auto const& type = typeid(std::string);
     int entry = 1;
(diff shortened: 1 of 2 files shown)


aolivier23 (dismissed):
Looks good to me.  Thank you!

gemmeren (dismissed):

form/storage/storage_reader.cpp
```diff
@@ -24,20 +24,21 @@ int StorageReader::getIndex(Token const& token,
...
-      m_read_containers.insert(
-        {key, createReadContainer(token.technology(), token.containerName())});
-      cont = m_read_containers.find(key);
-      for (auto const& [key, value] :
-           settings.getContainerTable(token.technology(), token.containerName()))
-        cont->second->setAttribute(key, value);
-      cont->second->setFile(file->second);
+      auto [cont_it, cont_inserted] = m_read_containers.try_emplace(
+          key, createReadContainer(token.technology(), token.containerName()));
+      if (cont_inserted) {
```
gemmeren: why conditional?

form/storage/storage_reader.cpp
```diff
@@ -60,19 +61,21 @@ void StorageReader::readContainer(Token const& token,
...
     }
-    m_read_containers.insert({key, createReadContainer(token.technology(), token.containerName())});
-    cont = m_read_containers.find(key);
-    cont->second->setFile(file->second);
-    for (auto const& [key, value] :
-         settings.getContainerTable(token.technology(), token.containerName()))
-      cont->second->setAttribute(key, value);
+    auto [cont_it, cont_inserted] = m_read_containers.try_emplace(
+        key, createReadContainer(token.technology(), token.containerName()));
+    if (cont_inserted) {
```
gemmeren: same here...

ALSO: there seems to be code duplication in `getIndex` and `readContainer`. I suspect copy/paste and it may be needed only in `getIndex` , can you please check and optimize?

wwuoneway:
## Update: Fixes applied

The original optimization introduced incorrect behavior that caused test failure:
- `try_emplace` unconditionally called `createFile()`/`createReadContainer`
- `StorageReader::getIndex()` and `readContainer()` logic was affected by premature initialization

This has been fixed by reverting to the safe `insert(...).first` pattern:
- Maintains efficient single lookup behavior
- Ensure constructors are only called when needed
- Preserves original lazy initialization semantics
- All tests pass

form/storage/storage_reader.cpp
wwuoneway: Good catch — this was a bug in the previous `try_emplace`-based implementation where constructors were invoked unconditionally.

In the current version, this has been fixed by switching to the safe `insert(...).first` pattern, ensuring proper lazy initialization.

See "Updated: Fixes applied" comment below.

form/storage/storage_reader.cpp
wwuoneway: Good point — both relate to the same underlying issue in the previous version.

In the earlier implementation, `getIndex` and `readContainer` shared duplicated and unsafe initialization logic (including the incorrect `try_emplace`-based pattern and conditional placement of `setFile` / `setAttribute`). This has been fixed in the current version by switching both functions to a consistent and safe lazy-initialization pattern using `insert(...).first`.

The two functions still serve different purposes:
- `getIndex`: builds and caches index mappings
- `readContainer`: performs actual data retrieval

Happy to refactor into a shared helper if further duplication becomes an issue in future changes.

form/storage/storage_reader.cpp
```diff
@@ -62,13 +63,15 @@ void StorageReader::readContainer(Token const& token,
   if (cont == m_read_containers.end()) {
     auto file = m_files.find(token.fileName());
```
aolivier23, line 64: Could we eliminate this `find()` by using the second element of `std::map::insert()`'s return value?

form/storage/storage_writer.cpp
```diff
@@ -29,9 +29,10 @@ void StorageWriter::createContainers(
       // Ensure the file exists
       auto file = m_files.find(plcmnt->fileName());
```
aolivier23, line 30: Can we eliminate this `find()` by using the second element of `std::map::insert()`'s return value?

form/storage/storage_reader.cpp
wwuoneway: (Continued) Good question — currently `readContainer()` does not strictly depend on a prior call to `getIndex()` . It performs its own lazy initialization of the file and container, so it can work independently as long as `token.id()`  is valid.
While in practice many call paths may go through getIndex()  first, this is not enforced by the API. Removing the initialization logic in `readContainer()`  would introduce an implicit precondition, which is currently neither documented or checked, and could make the code more fragile. That said, I agree there is some duplication. A cleaner approach would be to factor out, e.g., a shared helper. I'd prefer to keep the current version unchanged for this PR and address this in a follow-up PR.

form/storage/storage_reader.cpp
wwuoneway: Eliminating this `find()` would require unconditionally calling createFile() before knowing whether the file already exists. The current pattern is the correct lazy initialization approach with one `find()`.

form/storage/storage_writer.cpp
wwuoneway: same as above.

aolivier23 (approved):
Looks good to me.

PR #588, opened 2026-05-13
`clang-tidy`: resolve `performance-no-automatic-move`

greenc-FNAL:


greenc-FNAL:

phlex/core/index_router.cpp
```diff
@@ -175,7 +175,7 @@ namespace phlex::experimental {
     }
   }
 
-  data_cell_index_ptr index_router::route(data_cell_index_ptr const index, index_flushes flushes)
+  data_cell_index_ptr index_router::route(data_cell_index_ptr index, index_flushes flushes)
```
knoepfel: Hmm...the `index` parameter really should be `const`.  I can imagine passing in a `data_cell_index_ptr const&` object (which is a little strange, but not unheard of), or we might be able to change the call site to:

```diff
- return index_router_.route(index, flushes);
+ index_router_.route(*index, flushes);
+ return index;
```

and just pass in the dereferenced `index` pointer.  Another option is to just suppress the warning.

beojan: Does the `index` pointer get copied anywhere?

knoepfel: Actually, it does get copied.  Looks like the correct solution here is:

```suggestion
  data_cell_index_ptr index_router::route(data_cell_index_ptr const& index, index_flushes flushes)
```

knoepfel (approved):

PR #599, opened 2026-05-15
Form Test Writing Fundamental Types

aolivier23:
Adjusts `form_storage_test` to use FORM when testing that we can read fundamental types from a file.  We were using TTree directly before because FORM couldn't write certain fundamental types.  This also achieves:

- test that we can write each fundamental type
- fundamental type test should work with other FORM backends in the future
- simplifies our testing code a little


aolivier23:
@phlexbot format

gemmeren (approved):

wwuoneway (approved):
LGTM

PR #602, opened 2026-05-15
Move provider API away from misusing product_query for outputs

beojan:
Replaces #501

### DIFF
--- a/form/form_source.cpp
+++ b/form/form_source.cpp
@@ -84,14 +84,15 @@ PHLEX_REGISTER_PROVIDERS(s, config)
 
   // --- Register providers dynamically from config ---
   // FIXME: Prototype 0.1 -- types hardcoded as int.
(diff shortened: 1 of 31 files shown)


knoepfel (commented):
@beojan, this is looking quite good.  I noticed that you've chosen the explicit `experimental::algorithm_name::create("input")` instead of just `"input"`.  I think we should be able to support both patterns.  Probably involves an `output_product(...)` template that is constrained on `std::constructible_from<>` or `std::is_convertible<>` or ...?

My only other comment was in cleaning up the `match(...)` implementation, which you already addressed while I was assembling this review.

beojan:
As it happens, we already have an implicit constructor from string literals to `algorithm_name` so we didn't need to explicitly call `create`.

knoepfel (approved):

PR #603, opened 2026-05-15
Optimize some hot paths

knoepfel:
These are some optimizations that were suggested by Claude using macOS Instruments' profile data on the `many_events` test.

- Send the `algorithm_name` directly to product_store constructors instead of creating one from a std::string.
- Adjust the underlying container of `products` to be a `std::vector`.

### DIFF
--- a/phlex/core/consumer.hpp
+++ b/phlex/core/consumer.hpp
@@ -13,6 +13,7 @@ namespace phlex::experimental {
   public:
     consumer(algorithm_name name, std::vector<std::string> predicates);
 
+    algorithm_name const& name() const noexcept;
     std::string full_name() const;
     identifier const& plugin() const noexcept;
     identifier const& algorithm() const noexcept;
--- a/phlex/core/consumer.hpp
+++ b/phlex/core/consumer.hpp
@@ -14,7 +14,6 @@ namespace phlex::experimental {
     consumer(algorithm_name name, std::vector<std::string> predicates);
 
     algorithm_name const& name() const noexcept;
-    std::string full_name() const;
     identifier const& plugin() const noexcept;
     identifier const& algorithm() const noexcept;
     std::vector<std::string> const& when() const noexcept;
--- a/phlex/core/declared_fold.hpp
+++ b/phlex/core/declared_fold.hpp
@@ -126,7 +126,7 @@ namespace phlex::experimental {
     void emit_and_evict_if_done(data_cell_index_ptr const& fold_index)
     {
       if (auto counter = done_with(fold_index->hash())) {
-        auto parent = std::make_shared<product_store>(fold_index, this->full_name());
+        auto parent = std::make_shared<product_store>(fold_index, this->name());
         commit(parent);
         ++product_count_;
         tbb::flow::output_port<0>(fold_).try_put(
--- a/phlex/core/declared_fold.hpp
+++ b/phlex/core/declared_fold.hpp
@@ -65,7 +65,7 @@ namespace phlex::experimental {
     using function_t = typename AlgorithmBits::bound_type;
 
   public:
-    fold_node(algorithm_name name,
+    fold_node(algorithm_name algo_name,
               std::size_t concurrency,
               std::vector<std::string> predicates,
               tbb::flow::graph& g,
@@ -74,10 +74,10 @@ namespace phlex::experimental {
               product_queries input_products,
               std::vector<std::string> output,
               std::string partition) :
-      declared_fold{std::move(name), std::move(predicates), std::move(input_products)},
+      declared_fold{std::move(algo_name), std::move(predicates), std::move(input_products)},
       initializer_{std::move(initializer)},
       output_{
-        to_product_specifications(full_name(), std::move(output), make_type_ids<result_type>())},
+        to_product_specifications(name().full(), std::move(output), make_type_ids<result_type>())},
       partition_{std::move(partition)},
       flush_receiver_{g,
                       tbb::flow::unlimited,
@@ -92,7 +92,7 @@ namespace phlex::experimental {
                         return {};
                       }},
       join_{make_join_or_none<num_inputs>(
-        g, full_name(), layers())}, // FIXME: This should change to include result product!
+        g, name().full(), layers())}, // FIXME: This should change to include result product!
       fold_{g,
             concurrency,
             [this, ft = alg.release_algorithm()](messages_t<num_inputs> const& messages, auto&) {
@@ -126,7 +126,7 @@ namespace phlex::experimental {
     void emit_and_evict_if_done(data_cell_index_ptr const& fold_index)
     {
       if (auto counter = done_with(fold_index->hash())) {
-        auto parent = std::make_shared<product_store>(fold_index, this->name());
+        auto parent = std::make_shared<product_store>(fold_index, name());
         commit(parent);
         ++product_count_;
         tbb::flow::output_port<0>(fold_).try_put(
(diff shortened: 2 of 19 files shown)


phlex/core/consumer.hpp
```diff
@@ -13,6 +13,7 @@ namespace phlex::experimental {
   public:
     consumer(algorithm_name name, std::vector<std::string> predicates);
 
+    algorithm_name const& name() const noexcept;
     std::string full_name() const;
```
greenc-FNAL: Any reason these shouldn't be inline?

phlex/core/declared_fold.hpp
```diff
@@ -126,7 +126,7 @@ namespace phlex::experimental {
     void emit_and_evict_if_done(data_cell_index_ptr const& fold_index)
     {
       if (auto counter = done_with(fold_index->hash())) {
-        auto parent = std::make_shared<product_store>(fold_index, this->full_name());
+        auto parent = std::make_shared<product_store>(fold_index, this->name());
```
greenc-FNAL: Do we document why `name` _vs_ `full_name` here?

phlex/core/provider_node.cpp
```diff
@@ -26,10 +26,9 @@ namespace phlex::experimental {
                 auto new_product = std::invoke(ft, *index);
                 ++calls_;
 
-                products new_products;
+                products new_products{1uz};
```
greenc-FNAL, line 30: Why would the `new_products` container need to be initialized with a single default-constructed `product`?

phlex/model/products.hpp
```diff
@@ -78,8 +82,8 @@ namespace phlex::experimental {
     template <typename T>
     T const& get(product_specification const& spec) const
     {
-      auto it = products_.find(spec);
-      if (it == cend(products_)) {
+      auto it = std::ranges::find(products_, spec, [](auto const& p) { return p.first; });
```
greenc-FNAL: Should this be a public member function instead of a function-scope use of a lambda?

greenc-FNAL (dismissed):
Most of these comments reflect my unfamiliarity with the code rather than issues with the code itself.

Any chance of a test to mitigate https://app.codecov.io/gh/Framework-R-D/phlex/pull/603?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Framework-R-D?

phlex/core/consumer.hpp
knoepfel: It's historical: the other member functions were not inline, so I followed suit.  We should run an analysis on the code base some time and determine which functions should be inlined.

phlex/core/declared_fold.hpp
knoepfel: It's in the PR body:

> - Send the `algorithm_name` directly to `product_store` constructors instead of creating one from a `std::string`.

Probably, we should use different function names to connote the difference.

knoepfel: @copilot, do you have a suggestion for how to better name `consumer::full_name()` and `consumer::name()` based on their implementations?  Do not make any code changes at this time.

phlex/core/provider_node.cpp
knoepfel: This does not default-construct a product—it is the number used to reserve the capacity of the underlying vector that owns the data products.

phlex/model/products.hpp
knoepfel: I'm not sure that it should be public, but switching it to a member function makes sense (no reason to instantiate the lambda many times for each type `T`).  Will change.

knoepfel: The search has been moved to a private member function that is implemented in the `.cpp` file.

phlex/core/declared_fold.hpp
knoepfel: `name()` has been retained, but `full_name()` has been renamed `name_to_string()`.

knoepfel:
> Any chance of a test to mitigate https://app.codecov.io/gh/Framework-R-D/phlex/pull/603?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Framework-R-D?

The `contains()` interface was not used anywhere, so I've removed it.  It will be straightforward to add it back if we need it.

phlex/core/declared_fold.hpp
beojan: Is `name_to_string()` all that much better than `name().to_string()`? Though that is unfortunately currently `name().full()`.

phlex/core/declared_fold.hpp
```diff
@@ -76,8 +76,8 @@ namespace phlex::experimental {
               std::string partition) :
       declared_fold{std::move(name), std::move(predicates), std::move(input_products)},
       initializer_{std::move(initializer)},
-      output_{
-        to_product_specifications(full_name(), std::move(output), make_type_ids<result_type>())},
+      output_{to_product_specifications(
```
beojan: It looks like every time `to_product_specifications` is called an `algorithm_name` is converted to a string and passed in, then converted back to an `algorithm_name` in the body. We might as well pass the `algorithm_name` in here too.

knoepfel: Okay, we'll make the change.

phlex/core/declared_fold.hpp
knoepfel: @beojan, good point.  I'll remove the `consumer::name_to_string()` function in favor of `name().full()`.  Let's rename `algorithm_name::full()` to `algorithm_name::to_string()` in a separate PR.

knoepfel: @beojan, this change has been made with commit https://github.com/Framework-R-D/phlex/pull/603/commits/ffd10b9de601462b677cdf52632809dd0212ea1e.

phlex/core/declared_fold.hpp
knoepfel: This changes was made with commit https://github.com/Framework-R-D/phlex/pull/603/commits/1490a1d38cc81911c8af8924746f92823562f747.  However, it exposed an inconsistency in how `algorithm_name` can be constructed in multiple ways.  Commit https://github.com/Framework-R-D/phlex/pull/603/commits/5676f0f3f9e84723d8e7d65b3929183822913dba accommodates that inconsistency.  I don't think that needs to be sorted out for this PR, but we should address before v0.3.0 is released.

phlex/model/products.cpp
```diff
@@ -2,19 +2,30 @@
...
   products::const_iterator products::begin() const noexcept { return products_.begin(); }
   products::const_iterator products::end() const noexcept { return products_.end(); }
   products::size_type products::size() const noexcept { return products_.size(); }
   bool products::empty() const noexcept { return products_.empty(); }
 
+  product_base const* products::find_product(product_specification const& spec) const
+  {
+    auto it = std::ranges::find(products_, spec, [](auto const& p) { return p.first; });
+    if (it == products_.end()) {
+      throw std::runtime_error(fmt::format("No product exists with the name '{}'.", spec.full()));
```
beojan, line 24: With the "specification", I think.

phlex/core/provider_node.cpp
beojan: Could do with a comment. The behaviour is a little bit unexpected.

beojan (approved):
Couple of comments but they aren't critical.

phlex/model/products.cpp
knoepfel: Okay, will adjust it in a future PR.

phlex/core/provider_node.cpp
knoepfel: Okay, will come up with comment in another PR.

PR #614, opened 2026-05-26
Actions use other Phlex actions via pinned hash

greenc-FNAL:
This PR is mainly to satisfy CodeQL. It suffers from a minor hysteresis due to the presence of multiple reusable actions in the Phlex repository. Ideally this would be fixed by the usual convention of one action per dedicated repo, but that would occur an extra maintenance burden.

### DIFF
--- a/.github/actions/prepare-check-outputs/action.yaml
+++ b/.github/actions/prepare-check-outputs/action.yaml
@@ -44,11 +44,11 @@ runs:
     - name: Get PR Info
       if: github.event_name == 'issue_comment'
       id: get_pr
-      uses: Framework-R-D/phlex/.github/actions/get-pr-info@main
+      uses: Framework-R-D/phlex/.github/actions/get-pr-info@1b6b94e13ce2adbc2ba092d8d0fce6e11b252916 # actions-v0.1.0
 
     - name: Detect act environment
       id: detect_act
-      uses: Framework-R-D/phlex/.github/actions/detect-act-env@main
+      uses: Framework-R-D/phlex/.github/actions/detect-act-env@1b6b94e13ce2adbc2ba092d8d0fce6e11b252916 # actions-v0.1.0
 
     - name: Resolve outputs
       id: resolve
--- a/.github/actions/prepare-fix-outputs/action.yaml
+++ b/.github/actions/prepare-fix-outputs/action.yaml
@@ -26,7 +26,7 @@ runs:
     - name: Get PR Info
       if: github.event_name == 'issue_comment' && (inputs.ref == '' || inputs.repo == '')
       id: get_pr
-      uses: Framework-R-D/phlex/.github/actions/get-pr-info@main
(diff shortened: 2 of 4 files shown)


greenc-FNAL:

knoepfel (commented):
This seems problematic to me.  Suppose I want to alter the `get-pr-info` action.  It looks like I would need to:

1. Open a PR that makes changes to `actions/get-pr-info.yaml`
2. Pin the version `.../actions/get-pr-info@main` to test the changes
3. Merge that PR
4. Open another PR to pin the `get-pr-info` to the specific commit of the merged PR in step 3
5. Merge the other PR

Is there no other option?

greenc-FNAL:
> Is there no other option?

Put each reusable action in its own repository (this is standard practice, actually). Without a CI running on the action repository itself and/or with other restrictions, the need for PR-based branch protection for action repositories can be avoided. The workflows in other repositories using such actions can be tested by changing the workflows to use `<action>@main` on a branch, finalizing the PR in the client repository by updating the workflows to use the working, hash-pinned action only when the action has been fully tested and tagged. Workflows invoked by PR triggers are run from the PR branch, and workflows invoked by `workflow_dispatch` can be run from a specified branch; only workflows invoked by branch trigger or by PR comment are restricted to running from `main` IIRC.

knoepfel:
> > Is there no other option?
> 
> Put each reusable action in its own repository (this is standard practice, actually). Without a CI running on the action repository itself and/or with other restrictions, the need for PR-based branch protection for action repositories can be avoided. The workflows in other repositories using such actions can be tested by changing the workflows to use `<action>@main` on a branch, finalizing the PR in the client repository by updating the workflows to use the working, hash-pinned action only when the action has been fully tested and tagged. Workflows invoked by PR triggers are run from the PR branch, and workflows invoked by `workflow_dispatch` can be run from a specified branch; only workflows invoked by branch trigger or by PR comment are restricted to running from `main` IIRC.

Seems like a bit of work, but if we make the change now would that pay for itself down the road?  What's your opinion?

greenc-FNAL:
We'll be able to claim we're following best practice, and it will be easier for other repositories (e.g. Cetmodules) to use them. On the other hand, it will be easier for other repositories to use them. It would certainly be more straightforward to maintain the actions in a secure manner, but it might be more actual work (documentation and bookkeeping).

knoepfel:
> We'll be able to claim we're following best practice, and it will be easier for other repositories (e.g. Cetmodules) to use them. On the other hand, it will be easier for other repositories to use them. It would certainly be more straightforward to maintain the actions in a secure manner, but it might be more actual work (documentation and bookkeeping).

We should keep this in context—these actions are primarily for Phlex usage.  Refactoring them so that it's easier for us to adhere to security best practices is an argument in favor.  That does not mean that we need to document them right now beyond the scope of Phlex's needs.

I'd say do the refactoring and then this PR can be updated accordingly.

greenc-FNAL:
Superseded by #639

PR #615, opened 2026-05-26
Some cleanups per @beojan, etc.

knoepfel:
This PR implements a few cleanups:

- Renames the `full()` member functions to `to_string()`
- Removes an unnecessary `std::views::transform(...)`
- Explains a `1uz` constructor argument in `provider_node`
- Adjusts an exception message to use the name `specification` instead of `name`

### DIFF
--- a/form/form_module.cpp
+++ b/form/form_module.cpp
@@ -64,7 +64,7 @@ namespace {
       auto segment_id = store.index()->to_string();
 
       std::cout << "\n=== FormOutputModule::save_data_products ===\n";
-      std::cout << "Creator: " << creator.full() << "\n";
+      std::cout << "Creator: " << creator.to_string() << "\n";
       std::cout << "Segment ID: " << segment_id << "\n";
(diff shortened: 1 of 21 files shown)


greenc-FNAL:

beojan:
Ironically I think `algorithm_name` is the better name for the type. For instance, it's used in `observer_node` which isn't a creator.

This would be bike-shedding, but backing that change out looks like it would drastically reduce the size of this PR.

knoepfel:
> Ironically I think `algorithm_name` is the better name for the type. For instance, it's used in `observer_node` which isn't a creator.
> 
> This would be bike-shedding, but backing that change out looks like it would drastically reduce the size of this PR.

Oops.  The intention of the PR was to come up with better names.  And if `algorithm_name` is used by (e.g.) `observer_node`, then `creator_name` is not a good name.  Good catch.

knoepfel:
@beojan, can you re-review this PR?

beojan (approved):

PR #621, opened 2026-05-28
Make creator name optional

### LINKED ISSUE #623: Make creator name optional in `product_specification`

beojan:
Makes creator name optional in `product_selector`.

### DIFF
--- a/phlex/model/algorithm_name.hpp
+++ b/phlex/model/algorithm_name.hpp
@@ -28,7 +28,8 @@ namespace phlex::experimental {
     identifier const& algorithm() const noexcept { return algorithm_; }
 
     bool match(algorithm_name const& other) const;
-    auto operator<=>(algorithm_name const&) const = default;
+    auto operator<=>(algorithm_name const&) const noexcept = default;
+    bool operator==(algorithm_name const& rhs) const noexcept;
 
     static algorithm_name create(char const* spec);
     static algorithm_name create(std::string_view spec);
--- a/phlex/model/algorithm_name.hpp
+++ b/phlex/model/algorithm_name.hpp
@@ -9,8 +9,6 @@
 
 namespace phlex::experimental {
   class PHLEX_MODEL_EXPORT algorithm_name {
-    enum class specified_fields : std::uint8_t { neither, either, both };
-
   public:
     algorithm_name();
 
@@ -19,9 +17,7 @@ namespace phlex::experimental {
     algorithm_name(std::string const& spec);
     algorithm_name(std::string_view spec);
     // NOLINTEND(google-explicit-constructor)
-    algorithm_name(identifier plugin,
-                   identifier algorithm,
-                   specified_fields fields = specified_fields::both);
+    algorithm_name(identifier plugin, identifier algorithm);
 
     std::string to_string() const;
     identifier const& plugin() const noexcept { return plugin_; }
@@ -29,7 +25,6 @@ namespace phlex::experimental {
 
     bool match(algorithm_name const& other) const;
     auto operator<=>(algorithm_name const&) const noexcept = default;
-    bool operator==(algorithm_name const& rhs) const noexcept;
 
     static algorithm_name create(char const* spec);
     static algorithm_name create(std::string_view spec);
@@ -37,7 +32,6 @@ namespace phlex::experimental {
   private:
     identifier plugin_;
     identifier algorithm_;
-    specified_fields fields_{specified_fields::neither};
   };
 
 }
--- a/test/product_selector.cpp
+++ b/test/product_selector.cpp
@@ -34,4 +34,7 @@ TEST_CASE("Product name with data layer", "[data model]")
   CHECK(!product_selector{.creator = "1", .layer = "event", .suffix = "prod", .stage = "stage"_id}
            .match(product_selector{
              .creator = "1", .layer = "event", .suffix = "prod", .stage = "stage1"_id}));
+  // Here for no reason other than to satisfy codecov
+  CHECK(label == label);
+  CHECK((label <=> label) == std::strong_ordering::equal);
 }
--- a/test/product_selector.cpp
+++ b/test/product_selector.cpp
@@ -21,7 +21,7 @@ TEST_CASE("Product name with data layer", "[data model]")
   // Mismatched creator
   CHECK(!product_selector{.creator = "1", .layer = "event", .suffix = "prod"}.match(
     product_selector{.creator = "2", .layer = "event", .suffix = "prod"}));
-  // Missing creator matches anything
+  // Missing creator matches any creator
   CHECK(product_selector{.layer = "event", .suffix = "prod"}.match(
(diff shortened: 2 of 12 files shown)


greenc-FNAL:
## ✅ 7 CodeQL alerts resolved compared to main
- :white_check_mark: **Warning** [# 183](https://github.com/Framework-R-D/phlex/security/code-scanning/183) `actions/unpinned-tag` at `.github/actions/prepare-check-outputs/action.yaml:51:13` — Unpinned 3rd party Action 'action.yaml' step Uses Step: detect_act uses 'Framework-R-D/phlex/.github/actions/detect-act-env' with ref 'main', not a pinned commit hash
- :white_check_mark: **Warning** [# 184](https://github.com/Framework-R-D/phlex/security/code-scanning/184) `actions/unpinned-tag` at `.github/actions/prepare-fix-outputs/action.yaml:29:13` — Unpinned 3rd party Action 'action.yaml' step Uses Step: get_pr uses 'Framework-R-D/phlex/.github/actions/get-pr-info' with ref 'main', not a pinned commit hash
- :white_check_mark: **Warning** [# 185](https://github.com/Framework-R-D/phlex/security/code-scanning/185) `actions/unpinned-tag` at `.github/actions/run-change-detection/action.yaml:62:13` — Unpinned 3rd party Action 'action.yaml' step Uses Step: filter uses 'Framework-R-D/phlex/.github/actions/detect-relevant-changes' with ref 'main', not a pinned commit hash
- :white_check_mark: **Warning** [# 186](https://github.com/Framework-R-D/phlex/security/code-scanning/186) `actions/unpinned-tag` at `.github/actions/workflow-setup/action.yaml:68:13` — Unpinned 3rd party Action 'action.yaml' step Uses Step: prepare_check uses 'Framework-R-D/phlex/.github/actions/prepare-check-outputs' with ref 'main', not a pinned commit hash
- :white_check_mark: **Warning** [# 187](https://github.com/Framework-R-D/phlex/security/code-scanning/187) `actions/unpinned-tag` at `.github/actions/workflow-setup/action.yaml:79:13` — Unpinned 3rd party Action 'action.yaml' step Uses Step: prepare_fix uses 'Framework-R-D/phlex/.github/actions/prepare-fix-outputs' with ref 'main', not a pinned commit hash
- :white_check_mark: **Warning** [# 188](https://github.com/Framework-R-D/phlex/security/code-scanning/188) `actions/unpinned-tag` at `.github/actions/workflow-setup/action.yaml:128:13` — Unpinned 3rd party Action 'action.yaml' step Uses Step: detect uses 'Framework-R-D/phlex/.github/actions/run-change-detection' with ref 'main', not a pinned commit hash

Review the [full CodeQL report](https://github.com/Framework-R-D/phlex/security/code-scanning?query=pr%3A621) for details.

test/product_selector.cpp
```diff
@@ -24,6 +21,9 @@ TEST_CASE("Product name with data layer", "[data model]")
   // Mismatched creator
   CHECK(!product_selector{.creator = "1", .layer = "event", .suffix = "prod"}.match(
     product_selector{.creator = "2", .layer = "event", .suffix = "prod"}));
+  // Missing creator matches anything
```
knoepfel: I'd adjust this comment to "Missing creator matches any creator".

test/product_selector.cpp
```diff
@@ -7,9 +7,6 @@ using namespace phlex;
 
 TEST_CASE("Empty specifications", "[data model]")
 {
-  CHECK_THROWS_WITH(
-    (product_selector{.creator = "", .layer = "layer"}),
-    Catch::Matchers::ContainsSubstring("Cannot specify product with empty creator name."));
```
knoepfel, line 12: I think we still want this check—specifying `.creator = ""` is different than omitting the creator specification.

phlex/model/algorithm_name.hpp
```diff
@@ -28,7 +28,8 @@ namespace phlex::experimental {
     identifier const& algorithm() const noexcept { return algorithm_; }
 
     bool match(algorithm_name const& other) const;
-    auto operator<=>(algorithm_name const&) const = default;
+    auto operator<=>(algorithm_name const&) const noexcept = default;
+    bool operator==(algorithm_name const& rhs) const noexcept;
```
knoepfel: Do we use these comparison operators?  Providing both `operator<=>` and `operator==` is atypical, so that makes me wonder why we need both.  If we didn't use a `specified_fields` data member, then perhaps we wouldn't need separate operators.

One option: when constructing `algorithm_name`, the algorithm identifier is always required (a string with no colon specifies the algorithm name and not the plugin name).  But maybe this makes it impractical for folks who want to specify a plugin name in their product selector and don't necessarily know the algorithm identifier.

knoepfel (commented):
@beojan, this PR Is looking quite good.  Just a few comments below.  Any chance we can increase the [code coverage in `algorithm_name`](https://app.codecov.io/gh/Framework-R-D/phlex/pull/621?src=pr&el=tree&filepath=phlex%2Fmodel%2Falgorithm_name.cpp&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Framework-R-D#diff-cGhsZXgvbW9kZWwvYWxnb3JpdGhtX25hbWUuY3Bw)?

test/product_selector.cpp
beojan: This won't throw an error anymore because it's setting an `optional<identifier>`.

phlex/model/algorithm_name.hpp
beojan: The creator name in a product selector is a plain (optional) identifier which can match either the plugin name or the algorithm name. I could see what happens if I get rid of the `specified_fields` functionality, I suspect `execution_counts` might be where we run into trouble.

beojan: Didn't break anything other than the explicit test of this functionality.

test/product_selector.cpp
beojan: Done

test/product_selector.cpp
knoepfel: With this PR, it is no longer required to specify the creator.  But _when_ the creator is specified, I'm not sure we should allow the empty string.  If the empty string is specified, it will almost certainly not resolve to any known creator, and an exception will be thrown later on.  The only question is whether we make that check here or later.

beojan: Actually, because a name with no colon leaves the plugin unset, and a specified creator name can match plugin or algorithm, setting it to the empty string would match any algorithm with no plugin name.

I'm not sure if that's the behaviour we want. It seems like something that could be useful, but this might not be the way to express it.

knoepfel: Yeah, probably not the way we want to express that behavior.  I think, for now, we should forbid the specification of empty strings for `.creator`.  If that's tricky to do, then we can come up with something else.

beojan: OK, done.

knoepfel (approved):

PR #704, opened 2026-07-09
Layer checking as products pass over edges

beojan:
As discussed, layer checking happens dynamically as products pass over edges. A product selector without a layer specification would (in theory) accept products from any layer and place outputs appropriately so the layer a product belongs to need not be constant between iterations.

Layer specifications remain mandatory in selectors for now. Ironically until now they weren't actually being checked at all except when matching to products from providers.

-----------

- **Code**
  - Add dynamic layer-check nodes to `products_consumer`.
  - Forward a product only when its `layer_name()` matches the selector.
  - Store and expose the owning TBB graph in all declared node types.
  - Replace `find_producer()` with `find_producers()`.
  - Allow multiple producer matches when candidates differ only by layer.
  - Connect consumers to every matching producer.
  - Keep unmatched products for provider resolution and error reporting.
  - Build filter ports from consumer input selectors.

- **Tests**
  - Add coverage for selecting products by layer only.
  - Verify that the layer-only transform runs once.

- **Validation**
  - Review finding counts: unavailable from the supplied evidence.
  - Test execution results: unavailable from the supplied evidence.

### DIFF
--- a/phlex/core/make_computational_edges.cpp
+++ b/phlex/core/make_computational_edges.cpp
@@ -154,7 +154,7 @@ namespace phlex::detail {
             continue;
           }
 
-          for (auto* producer : producer_ports) {
+          for (auto const* producer : producer_ports) {
             make_edge(*producer->output_port, *receiver_port);
           }
         }
--- a/phlex/core/make_computational_edges.cpp
+++ b/phlex/core/make_computational_edges.cpp
@@ -147,14 +147,16 @@ namespace phlex::detail {
 
         for (auto const& query : node->input()) {
           auto* receiver_port = collector ? collector : &node->port(query);
-          auto const* producer = producers.find_producer(query, node->name());
-          if (not producer) {
+          auto producer_ports = producers.find_producers(query, node->name());
+          if (producer_ports.empty()) {
             // Is there a way to detect mis-specified product dependencies?
             result[node_name].push_back({query, receiver_port});
             continue;
           }
 
-          make_edge(*producer->output_port, *receiver_port);
+          for (auto* producer : producer_ports) {
+            make_edge(*producer->output_port, *receiver_port);
+          }
         }
       }
       return result;
--- a/phlex/core/producer_catalog.cpp
+++ b/phlex/core/producer_catalog.cpp
@@ -1,5 +1,6 @@
 #include "phlex/core/producer_catalog.hpp"
 #include "phlex/utilities/bulleted_list.hpp"
+#include "phlex/utilities/hashing.hpp"
 
 #include "fmt/format.h"
 #include "fmt/ranges.h"
@@ -7,13 +8,13 @@
 #include <ranges>
 
 namespace phlex::detail {
-  producer_catalog::named_output_port const* producer_catalog::find_producer(
+  std::vector<producer_catalog::named_output_port const*> producer_catalog::find_producers(
     product_selector const& query, phlex::experimental::algorithm_name const& consumer_name) const
   {
     if (producers_.empty()) {
       spdlog::debug("No producers found. Skipping and assuming {} comes from a provider.",
                     query.to_string());
-      return nullptr;
+      return {};
     }
     // Now the only way b == e is if we have a suffix and nothing creates matching products
     auto [b, e] = query.suffix.has_value() ? producers_.equal_range(*query.suffix)
@@ -22,9 +23,18 @@ namespace phlex::detail {
       spdlog::debug(
         "Failed to find an algorithm that creates {} products. Assuming it comes from a provider",
         query.suffix.value_or("*"_id));
-      return nullptr;
+      return {};
     }
-    std::map<std::string, named_output_port const*> candidates;
+    std::vector<named_output_port const*> candidates;
+
+    // Don't really want to copy these fields so we'll use hashes and store a pointer to one copy
+    std::map<std::uint64_t, experimental::identifier const*> suffixes;
+    if (query.suffix.has_value()) {
+      suffixes.emplace(query.suffix->hash(), &*query.suffix);
+    }
+    std::map<std::uint64_t, experimental::algorithm_name const*> creators;
+    std::map<std::uint64_t, type_id const*> types;
+
     for (auto const& [key, producer] : std::ranges::subrange{b, e}) {
       // Prevent self-edges
       if (producer.node == consumer_name) {
@@ -59,7 +69,14 @@ namespace phlex::detail {
                          query.type.exact_name(),
                          producer.type.exact_name());
           }
-          candidates.emplace(producer.node.to_string(), &producer);
+          candidates.push_back(&producer);
+          if (!query.suffix.has_value()) {
+            suffixes.emplace(key.hash(), &key);
+          }
+          creators.emplace(
+            phlex::detail::hash(producer.node.plugin().hash(), producer.node.algorithm().hash()),
+            &producer.node);
+          types.emplace(hash_value(producer.type), &producer.type);
         }
       } else {
         spdlog::debug("Creator name mismatch between ({}) and {}",
@@ -72,17 +89,41 @@ namespace phlex::detail {
       spdlog::debug(
         "Cannot identify product matching the query {}. Assuming it comes from a provider.",
         query.to_string());
-      return nullptr;
     }
 
     if (candidates.size() > 1ull) {
-      std::string msg = fmt::format("More than one candidate matches the query {}: \n{}\n",
+      static auto const port_to_node =
+        [](named_output_port const* p) -> experimental::algorithm_name const& { return p->node; };
+      static auto const deref_view =
+        std::views::transform([]<typename T>(T const* p) -> T const& { return *p; });
+      std::string msg = fmt::format("More than one candidate matches the query {}: \n{}",
                                     query.to_string(),
-                                    bulleted_list(std::views::keys(candidates),
+                                    bulleted_list(std::views::transform(candidates, port_to_node),
                                                   /*indent=*/1));
-      throw std::runtime_error(msg);
+      if (suffixes.size() == 1 && creators.size() == 1 && types.size() == 1) {
+        spdlog::info(msg);
+        spdlog::info("This is permitted -- layers may differ");
+        return candidates;
+      }
+      spdlog::error(msg);
+
+      if (suffixes.size() > 1) {
+        spdlog::error("Not permitted -- distinguishable by suffix {}",
+                      suffixes | std::views::values | deref_view);
+      }
+      if (creators.size() > 1) {
+        spdlog::error("Not permitted -- distinguishable by creator {}",
+                      creators | std::views::values |
+                        std::views::transform(
+                          [](experimental::algorithm_name const* p) { return p->to_string(); }));
+      }
+      if (types.size() > 1) {
+        spdlog::error("Not permitted -- distinguishable by type {}",
+                      types | std::views::values | deref_view);
+      }
+      throw std::runtime_error("Multiple products in candidate set -- see errors");
     }
 
-    return candidates.begin()->second;
+    return candidates;
   }
 }
--- a/phlex/core/producer_catalog.cpp
+++ b/phlex/core/producer_catalog.cpp
@@ -11,6 +11,13 @@ namespace phlex::detail {
   std::vector<producer_catalog::named_output_port const*> producer_catalog::find_producers(
     product_selector const& query, phlex::experimental::algorithm_name const& consumer_name) const
   {
+    // Will need an update when we have a way to set the current stage name
+    if (query.stage.has_value() && query.stage.value() != "CURRENT"_idq) {
+      spdlog::debug(
+        "{} requires a stage other than the current one. Assuming it comes from a provider.",
+        query.to_string());
+      return {};
+    }
     if (producers_.empty()) {
       spdlog::debug("No producers found. Skipping and assuming {} comes from a provider.",
                     query.to_string());
--- a/phlex/core/products_consumer.cpp
+++ b/phlex/core/products_consumer.cpp
@@ -1,4 +1,5 @@
 #include "phlex/core/products_consumer.hpp"
+#include <spdlog/spdlog.h>
 
 namespace {
   std::vector<phlex::experimental::identifier> layers_from(phlex::product_selectors const& queries)
@@ -29,7 +30,20 @@ namespace phlex::detail {
 
   tbb::flow::receiver<message>& products_consumer::port(product_selector const& input_product)
   {
-    return port_for(input_product);
+    // Everything has a layer for now, so everything needs this
+    auto& next = port_for(input_product);
+
+    auto& layer_check = layer_checkers_.emplace_back(std::make_unique<layer_check_node_t>(
+      graph(),
+      tbb::flow::unlimited,
+      [&layer = static_cast<experimental::identifier const&>(input_product.layer)](
+        message const& msg, auto& output) {
+        if (msg.store->layer_name() == layer) {
+          std::get<0>(output).try_put(msg);
+        }
+      }));
+    make_edge(tbb::flow::output_port<0>(*layer_check), next);
+    return *layer_check;
   }
 
   product_selectors const& products_consumer::input() const noexcept { return input_products_; }
--- a/phlex/core/products_consumer.hpp
+++ b/phlex/core/products_consumer.hpp
@@ -18,6 +18,8 @@
 
 namespace phlex::detail {
   class PHLEX_CORE_EXPORT products_consumer : public consumer {
+    using layer_check_node_t = tbb::flow::multifunction_node<message, message_tuple<1UZ>>;
+
   public:
     products_consumer(phlex::experimental::algorithm_name name,
                       std::vector<std::string> predicates,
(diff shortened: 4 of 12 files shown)


beojan:
@codex review

test/product_selecting.cpp
```diff
@@ -99,4 +99,23 @@ TEST_CASE("Querying products in different ways", "[graph]")
     g.execute();
     CHECK(g.execution_count("creator_and_layer_after_transform") == num_events);
   }
+
+  SECTION("Products from this job, using layer only")
+  {
+    // create a product to be found
+    g.fold(
+       "duplicate_temperature",
```
knoepfel, line 110: This should trigger an exception before executing the graph: there's already a transform algorithm named `"duplicate_temperature"` earlier in the file.

phlex/core/producer_catalog.cpp
```diff
@@ -72,17 +89,41 @@ namespace phlex::detail {
...
+      std::string msg = fmt::format("More than one candidate matches the query {}: \n{}",
                                     query.to_string(),
-                                    bulleted_list(std::views::keys(candidates),
+                                    bulleted_list(std::views::transform(candidates, port_to_node),
                                                   /*indent=*/1));
-      throw std::runtime_error(msg);
+      if (suffixes.size() == 1 && creators.size() == 1 && types.size() == 1) {
+        spdlog::info(msg);
+        spdlog::info("This is permitted -- layers may differ");
+        return candidates;
```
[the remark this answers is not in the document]
beojan: @codex Can you suggest how I might detect this situation?

phlex/core/make_computational_edges.cpp
```diff
@@ -147,14 +147,16 @@ namespace phlex::detail {
 
         for (auto const& query : node->input()) {
           auto* receiver_port = collector ? collector : &node->port(query);
-          auto producer = producers.find_producer(query, node->name());
-          if (not producer) {
+          auto producer_ports = producers.find_producers(query, node->name());
+          if (producer_ports.empty()) {
             // Is there a way to detect mis-specified product dependencies?
             result[node_name].push_back({query, receiver_port});
```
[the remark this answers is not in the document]
beojan: @codex Strictly, those products are also distinguishable by stage name so a stage name should be provided. Our semantics are that where no stage name is provided that ambiguity is resolved by preferring the product in the current stage.

Nevertheless, can you suggest a way to detect this ambiguity where it actually exists?

test/product_selecting.cpp
```diff
@@ -99,4 +99,23 @@ TEST_CASE("Querying products in different ways", "[graph]")
     g.execute();
     CHECK(g.execution_count("creator_and_layer_after_transform") == num_events);
   }
+
+  SECTION("Products from this job, using layer only")
```
knoepfel, line 106: Does the section above (i.e., `SECTION("Layer alone, distinguished by type")`) not already cover this case?

knoepfel (commented):
@beojan, thanks for the PR.  I want to check a few things first with you:

1. Multiple incoming edges to one input port are allowed so long as the source nodes of those incoming edges differ only in the layer.
2. Assuming (1) is true, that means the only ports that can have multiple incoming edges are those input ports that are connected to provider nodes.  All other input ports are connected to nodes that have differing names (where I assume a provider-node name cannot clash with a computational-node name)
3. Assuming (2) is true, the only layer-checking that is necessary is the following situation: Node B wants to receive only layer `p` data from Node A, whereas Node A receives both layer `p` and layer `q` data from corresponding upstream provider nodes.

Does that accord with your understanding?

test/product_selecting.cpp
beojan: It certainly doesn't appear to.

beojan:
1 is how I understand things, 2 evidently seems to be untrue. It appears nodes can share names so long as you don't try to wire them directly to each other.

I'm less sure 1 *should* be true. If you have one path for generating and simulating MC and another for collecting real data, it doesn't seem to be that useful to say "Phlex can handle both in one stream and deal with the layer hierarchies differing, but only if creator names and product suffixes are identical in both paths".

beojan:
3 also doesn't follow from 2 and 1. If you write a selector that matches some product created by a previous node in every way except the layer is wrong, it shouldn't match anything. Without layer checking, it'll still match.

test/product_selecting.cpp
beojan: No, the section above specifies the layer only because layer is a mandatory field. The product lookup is done by type.

knoepfel: But with your changes, wouldn't the lookup then also be done by layer as well?

I'm trying to determine how the construction you've added is different than what's already there...and it's not clear to me how it is.

knoepfel:
> I'm less sure 1 _should_ be true. If you have one path for generating and simulating MC and another for collecting real data, it doesn't seem to be that useful to say "Phlex can handle both in one stream and deal with the layer hierarchies differing, but only if creator names and product suffixes are identical in both paths".

This is a fair point, and I now realize the issue is more nuanced than I thought.  If all of the product-selector fields need to be identical except for the layer, then that means that, for the same stage, the algorithm names need to be identical.  And I argue that should be an error—nodes in the same stage are currently distinguishable (to the user) via the algorithm name, and thus they should be unique.  

So...that means we have a data-product provenance issue we need to disentangle.  Will it be a problem for users if not only the data layers differ upon node execution, but also the originating provenance (creator names, stage names, etc.) of the input products?  Maybe we can talk about this on Monday?

test/product_selecting.cpp
beojan: The selector requires layer to be `event` and type to be `string` (`LIST char`). Before and after my change, the product is uniquely identified by the type alone (i.e. there's only one node producing `string` products).

In the new test, the layer name is required to disambiguate, since there are two `duplicate_temperature` nodes producing `temperature` products of type `double` in the current stage.

phlex/core/make_computational_edges.cpp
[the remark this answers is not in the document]
beojan: @codex Partially resolved by bailing early if a stage name other than "CURRENT" is requested.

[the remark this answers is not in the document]
beojan: Local Codex suggests this is unfortunately not resolvable.

phlex/core/producer_catalog.cpp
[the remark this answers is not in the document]
beojan: In the general (dynamic) case, this appears to be undetectable.

phlex/core/make_computational_edges.cpp
```diff
@@ -147,14 +147,16 @@ namespace phlex::detail {
...
             continue;
           }
 
-          make_edge(*producer->output_port, *receiver_port);
+          for (auto* producer : producer_ports) {
+            make_edge(*producer->output_port, *receiver_port);
```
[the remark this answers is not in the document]
beojan: Addressed

beojan:
@coderabbitai full review

phlex/core/make_computational_edges.cpp
```diff
@@ -163,14 +163,16 @@ namespace phlex::detail {
...
+          if (producer_ports.empty()) {
             // Is there a way to detect mis-specified product dependencies?
             result[node_name].push_back({.input_product = query, .port = receiver_port});
             continue;
           }
 
-          make_edge(*producer->output_port, *receiver_port);
+          for (auto const* producer : producer_ports) {
+            make_edge(*producer->output_port, *receiver_port);
+          }
```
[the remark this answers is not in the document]
beojan: Addressed.

phlex/core/producer_catalog.cpp
```diff
@@ -36,53 +156,19 @@ namespace phlex::detail {
...
-      std::string msg = fmt::format("More than one candidate matches the query {}: \n{}\n",
-                                    query.to_string(),
-                                    bulleted_list(std::views::keys(candidates),
-                                                  /*indent=*/1));
-      throw std::runtime_error(msg);
-    }
-
-    return candidates.begin()->second;
+    check_postconditions(query, candidates, suffixes, creators, types);
+    return candidates;
```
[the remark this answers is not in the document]
beojan: Addressed

phlex/core/products_consumer.cpp
```diff
@@ -64,7 +65,24 @@ namespace phlex::detail {
...
-    return port_for(input_product);
+    auto& next = port_for(input_product);
+
+    // If input_product doesn't have a layer, it must be for a node that allows layer omission
+    if (input_product.layer) {
+      auto& layer_check = layer_checkers_.emplace_back(std::make_unique<layer_check_node_t>(
+        graph(),
+        tbb::flow::unlimited,
+        [&layer = static_cast<experimental::identifier const&>(input_product.layer)](
+          message const& msg, auto& output) {
```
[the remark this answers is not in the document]
beojan: The identifier always comes from a persistent list of selectors.

phlex/core/products_consumer.hpp
```diff
@@ -46,9 +48,10 @@ namespace phlex::detail {
 
   private:
     virtual tbb::flow::receiver<message>& port_for(product_selector const& input_product) = 0;
-
+    virtual tbb::flow::graph& graph() const = 0;
```
knoepfel: Recommend storing a reference/pointer to the `tbb::flow::graph` in the base class itself (should be passed in as a constructor argument) instead of in the derived classes.

test/filter_impl_test.cpp
knoepfel, line 1: Why the changes in this file?

phlex/core/detail/filter_impl.cpp
```diff
@@ -71,10 +92,8 @@ namespace phlex::detail {
       a->second.resize(nargs_);
     }
     auto& elem = a->second;
-    if (nargs_ == 1ull) {
-      // We do not check that the product is in the store if only one argument is
-      // forwarded.  This enables us to forward arguments for regular nodes and also
-      // output nodes, which do not take individual data products.
+    if (is_output_only(*input_products_)) {
```
knoepfel, line 79: Suggest something simpler like:

```cpp
  if (*input_products_ == for_output_only()) { ... }

File: phlex/core/producer_catalog.cpp [hunk 11]
```diff
@@ -1,30 +1,182 @@
...
+    }
+    throw std::runtime_error("Multiple products in candidate set -- see errors");
+  }
+}
+
+namespace phlex::detail {
+  std::vector<producer_catalog::named_output_port const*> producer_catalog::find_producers(
+    product_selector const& query, phlex::experimental::algorithm_name const& consumer_name) const
+  {
+    if (preconditions_violated(query, producers_)) {
```
knoepfel: Consider reversing the sign of the check:

```cpp
if (not preconditions_satisfied(query, producers_)) { ... }
```

Also, the term "preconditions" may need some refinement.  Generally, a function's preconditions must be satisfied for a function to meaningfully complete.  However, if `preconditions_violated(...)` returns true, the function returns an empty vector.  That's a successful function completion.  I would expect (e.g.) an exception throw if the preconditions were violated.

phlex/core/producer_catalog.cpp
```diff
@@ -1,30 +1,182 @@
...
+    };
+    static auto const deref_view =
+      std::views::transform([]<typename T>(T const* p) -> T const& { return *p; });
+    std::string msg = fmt::format("More than one candidate matches the query {}: \n{}",
+                                  query.to_string(),
+                                  bulleted_list(std::views::transform(candidates, port_to_node),
+                                                /*indent=*/1));
+    if (suffixes.size() == 1 && creators.size() == 1 && types.size() == 1) {
+      spdlog::info(msg);
+      spdlog::info("This is permitted -- layers may differ");
```
knoepfel: I might switch these to `spdlog::debug`.

phlex/core/producer_catalog.cpp
```diff
@@ -1,30 +1,182 @@
...
+      spdlog::error("Not permitted -- distinguishable by creator {}",
+                    creators | std::views::values |
+                      std::views::transform(
+                        [](experimental::algorithm_name const* p) { return p->to_string(); }));
+    }
+    if (types.size() > 1) {
+      spdlog::error("Not permitted -- distinguishable by type {}",
+                    types | std::views::values | deref_view);
+    }
+    throw std::runtime_error("Multiple products in candidate set -- see errors");
```
knoepfel: I would place this throw closer to the checks on the size of candidates up above.

phlex/core/producer_catalog.cpp
```diff
@@ -1,30 +1,182 @@
...
+    }
+    if (candidates.size() == 1ull) {
+      return;
+    }
+
+    static auto const port_to_node = [](named_output_port const* p) -> algorithm_name const& {
+      return p->node;
+    };
+    static auto const deref_view =
+      std::views::transform([]<typename T>(T const* p) -> T const& { return *p; });
```
knoepfel, line 125:
```suggestion
      std::views::transform([](auto const* p) -> auto const& { return *p; });
```

phlex/core/producer_catalog.cpp
```diff
@@ -1,30 +1,182 @@
...
+      spdlog::info(msg);
+      spdlog::info("This is permitted -- layers may differ");
+      return;
+    }
+
+    spdlog::error(msg);
+
+    if (suffixes.size() > 1) {
+      spdlog::error("Not permitted -- distinguishable by suffix {}",
+                    suffixes | std::views::values | deref_view);
```
knoepfel, line 140: Is the error message suggesting that the candidates _should be_ distinguished by suffix?

phlex/core/producer_catalog.cpp
```diff
@@ -1,30 +1,182 @@
...
         query.suffix.value_or("*"_id));
-      return nullptr;
+      return {};
+    }
+    std::vector<named_output_port const*> candidates;
+
+    // Don't really want to copy these fields so we'll use hashes and store a pointer to one copy
+    std::map<std::uint64_t, experimental::identifier const*> suffixes;
+    if (query.suffix.has_value()) {
+      add_to_tracker(suffixes, query.suffix->hash(), *query.suffix, "identifiers");
```
knoepfel, line 177: We should discuss whether we want to worry about hash collisions.  This discussion does not need to hold up this PR, but this is not the only place where we worry about hashed values, and we should be intentional about how to handle them.

knoepfel (commented):
@beojan, thanks for the PR.  Some comments below.

test/filter_impl_test.cpp
beojan: Without these changes the products are actually placed into the `job` layer. Now that the data map actually checks the layer that caused a failure.

phlex/core/producer_catalog.cpp
beojan: `should_skip_providers`?

beojan: Actually `providers_only`, I suppose.

phlex/core/producer_catalog.cpp
beojan: It's saying there are multiple matches that could be distinguished by suffix, and therefore a suffix must be provided.

phlex/core/producer_catalog.cpp
beojan: I don't think we really need to worry, but I didn't feel like arguing with CodeRabbit.

phlex/core/producer_catalog.cpp
beojan: Done

phlex/core/detail/filter_impl.cpp
beojan: Done

phlex/core/products_consumer.hpp
beojan: Done

phlex/core/products_consumer.cpp
```diff
@@ -64,7 +67,24 @@ namespace phlex::detail {
 
   tbb::flow::receiver<message>& products_consumer::port(product_selector const& input_product)
   {
-    return port_for(input_product);
+    auto& next = port_for(input_product);
+
+    // If input_product doesn't have a layer, it must be for a node that allows layer omission
+    if (input_product.layer) {
+      auto& layer_check = layer_checkers_.emplace_back(std::make_unique<layer_check_node_t>(
+        graph(),
```
knoepfel: Suggest just using `graph_`.

phlex/core/products_consumer.hpp
```diff
@@ -46,9 +50,11 @@ namespace phlex::detail {
 
   private:
     virtual tbb::flow::receiver<message>& port_for(product_selector const& input_product) = 0;
-
+    tbb::flow::graph& graph() const { return graph_; }
```
knoepfel: Suggest removing this accessor and using `graph_` directly where necessary.

phlex/core/producer_catalog.cpp
knoepfel: @coderabbit, can you create an issue that reminds us to discuss the issue of hash collisions?

phlex/core/producer_catalog.cpp
knoepfel: Maybe rephrase the error message then (e.g.):

```cpp
spdlog::error("The selector matches multiple products with different suffixes ({})."
              "Specify a suffix in the selector to disambiguate.", ...);
```

and similarly for the other error messages below this one.

test/filter_impl_test.cpp
knoepfel: Ah, got it.

knoepfel (commented):
Looking good, @beojan.  Just a few more comments.

phlex/core/products_consumer.hpp
beojan: Done

phlex/core/products_consumer.cpp
beojan: Done

phlex/core/producer_catalog.cpp
beojan: Done

knoepfel (approved):

PR #718, opened 2026-07-15
Phlex v0.3.2

### LINKED ISSUE #721: Products published by phlex::source subclasses never reach output modules
Output modules (e.g. `form_module`) receive stores from explicit providers and from producers, but never from `phlex::source` subclasses that publish products via `create_providers`.

Cause: `edges_to_outputs` wires provider and producer output ports to the output nodes:
https://github.com/Framework-R-D/phlex/blob/88ed89b2f6506b827f266344c37aa214992b672e/phlex/core/make_computational_edges.cpp#L175-L187, and it runs *before* `edges_from_implicit_providers` creates the implicit provider nodes from the registered sources (L204 vs L209). At that point `nodes.providers` only contains the explicit providers, so a source's products are only materialised for consumers that request them — output modules never see them.

Symptom: a workflow with `source -> transform -> form_module` writes the transform's product but silently omits the source's, even when it is listed in the module's `products` configuration. The FORM tests don't hit this because their event data comes from explicit provider plugins (e.g. `test/plugins/ij_source.cpp` uses `PHLEX_REGISTER_PROVIDERS`).

Downstream (SHiP) we work around it with a pass-through transform that consumes the source's product and re-emits it unchanged, at the cost of a copy per event.

Maybe the usecase for this is limited, but at least right now for prototyping, it is useful to be able to observe the output directly from the source.

Possibly interacts with the implicit-provider emplacement question in #667. A fix probably needs the output edges to be drawn (or re-drawn) after implicit providers exist, though only for products an output module actually persists, since creating implicit providers for *all* source products just to feed outputs would materialise data nobody else consumes.

knoepfel:
This PR reflects the Phlex v0.3.2 bug-fix release.  It consists of:

- Cherry-picked commits from `main`
- A fix to #721 
- An updated version number

### DIFF
--- a/.github/workflows/codeql-analysis.yaml
+++ b/.github/workflows/codeql-analysis.yaml
@@ -357,7 +357,8 @@ jobs:
 
       - name: Prepare PR comment data
         if: >-
-          github.event_name == 'pull_request' && steps.check_codeql.outputs.comment_path != ''
+          github.event_name == 'pull_request' && (steps.check_codeql.outputs.new_vs_base == 'true' ||
+          steps.check_codeql.outputs.fixed_vs_base == 'true') && steps.check_codeql.outputs.comment_path != ''
         env:
           COMMENT_PATH: ${{ steps.check_codeql.outputs.comment_path }}
         run: |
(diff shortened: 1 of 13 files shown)


aolivier23 (commented):
FORM changes look good to me.

aolivier23 (approved):
FORM changes look good to me.  Thank you.

greenc-FNAL (approved):
The new test is a regression test, yes? Failed before your fix, passes after?

knoepfel:
> The new test is a regression test, yes? Failed before your fix, passes after?

That's correct.

knoepfel:
@coderabbit, your comment at https://github.com/Framework-R-D/phlex/pull/718#pullrequestreview-4708060030 is incorrect.  The implicitly generated copy and move constructors require no template deduction and are thus favored over the forwarding constructor that does require template deduction.

PR #720, opened 2026-07-15
`clang-tidy`: resolve `modernize-use-using`

knoepfel:
- **Code quality**
  - Replaced legacy `typedef` declarations with modern C++ `using` aliases in Python plugin sources.
  - Updated aliases for platform integer types, `dcargs_t`, shared-pointer storage, and graph proxy types.
  - Preserved all underlying types and runtime behavior.


plugins/python/src/dyncall.hpp
```diff
@@ -27,8 +27,8 @@
 #if defined(__APPLE__) && defined(__MACH__)
 // This is a temporary workaround until we have a solution for handling translation of types
 // between C++ and Python.
-typedef long ph_long_t;
-typedef unsigned long ph_ulong_t;
+using ph_long_t = long;
+using ph_ulong_t = unsigned long;
 #else
 typedef std::int64_t ph_long_t;
 typedef std::uint64_t ph_ulong_t;
```
greenc-FNAL: In order to be correct for non-Mac builds, something like the following would be appropriate:

```suggestion
using ph_long_t = std::int64_t;
using ph_ulong_t = std::uint64_t;
```

greenc-FNAL (changes requested):

plugins/python/src/dyncall.hpp
knoepfel: Good catch, @greenc-FNAL.

greenc-FNAL (approved):

PR #727, opened 2026-07-16
`clang-tidy`: resolve `modernize-use-starts-ends-with`

knoepfel:
- **Code quality:** Replaced manual `std::string::compare` prefix checks with `std::string::starts_with` for `ndarray` and `list` collection-type detection.
- **Scope:** Updated converter insertion, Numba transform registration, and collection-output source registration paths without changing behavior or public APIs.


greenc-FNAL (changes requested):
Changes look fine, but we're a bit shy on coverage. Up to you whether you want to try to improve it or just force it through because there's no actual change in covered code _vs_ before.

knoepfel:
> Changes look fine, but we're a bit shy on coverage. Up to you whether you want to try to improve it or just force it through because there's no actual change in covered code _vs_ before.

Agreed that the coverage could be improved. The scope of the PR, however, is just the application of a targeted `clang-tidy` fix.  Any intentional adjustments to code coverage should arguably be part of a different PR.

greenc-FNAL (approved):

PR #745, opened 2026-07-21
`clang-tidy`: resolve `modernize-avoid-c-arrays`

knoepfel:
- **Code quality**
  - Replaced raw C-style character and method-definition arrays with `std::array` in string-literal and Python wrapper utilities.
  - Updated array access to use `.data()` where C APIs require raw pointers.
  - Modernized NumPy dimension storage with `std::array<npy_intp, 1>`.

- **Python compatibility**
  - Added conditional keyword-table handling for Python versions before and from 3.13, supporting the differing `PyArg_ParseTupleAndKeywords` requirements.
  - Preserved sentinel terminators in converted `PyMethodDef` tables.

- **String utilities**
  - Updated `string_literal` storage and construction to use `std::array<char, N>` with `std::ranges::copy`.
  - Explicitly returns the array’s character buffer when converting to `std::string_view`.


plugins/python/src/modulewrap.cpp
```diff
@@ -759,14 +760,24 @@ static PyObject* parse_args(PyObject* args,
   // any node. (The observer does not require outputs, but they still need to be
   // retrieved, not ignored, to issue an error message if an output is provided.)
 
-  static char kw0[] = "callable", kw1[] = "input_family", kw2[] = "output_product_suffixes",
-              kw3[] = "concurrency", kw4[] = "name";
-  // kwnames can be of type char const*[] once we mandate Python 3.13 or newer
+  // Python 3.13+ accepts const keyword-name tables. Older versions require mutable strings.
```
marcpaterno, line 763: I suggest we strongly encourage DUNE to officially adopt a Python version (and Python library version) dependency policy following the lead of Scientific Python Ecosystem Coordination (SPEC) policy 000 (https://scientific-python.org/specs/spec-0000). The reasons for doing so are spelled out in the SPEC 0 description, and elsewhere in the SPEC documentation.

Following those guidelines, we would drop support for Python 3.12 (and earlier) by our first release following October 2026. At the time we drop support for Python 3.12, this code should be simplified.

This comment should not block this PR.

marcpaterno (approved):

plugins/python/src/modulewrap.cpp
knoepfel: Thanks, @marcpaterno.  Can you create a GitHub discussion at https://github.com/orgs/Framework-R-D/discussions that discusses this proposal?

PR #749, opened 2026-07-22
Fix FORM's RNTuple backend build against ROOT 6.40

### LINKED ISSUE #712: FORM's RNTuple storage backend does not compile against ROOT ≥ 6.40 (RRawPtrWriteEntry namespace move)
Building with -DFORM_USE_RNTUPLE_STORAGE=ON fails against ROOT 6.40 (observed with conda-forge root_base 6.40.02; affects v0.3.0, v0.3.1, and current main).

ROOT 6.40 moved RRawPtrWriteEntry out of ROOT::Experimental::Detail into ROOT::Detail, but the forward declaration and the m_entry member in root_rntuple_write_container.hpp  still use the pre-6.40 namespace:
https://github.com/Framework-R-D/phlex/blob/88ed89b2f6506b827f266344c37aa214992b672e/form/root_storage/root_rntuple_write_container.hpp#L17-L19

I added a version-conditional patch (#if ROOT_VERSION_CODE >= ROOT_VERSION(6, 40, 0) around the forward declaration plus a using alias for the member) in the conda-forge phlex-feedstock (conda-forge/phlex-feedstock#8) so ROOT 6.38 keeps building. Happy to upstream this patch.

Related: since FORM_USE_RNTUPLE_STORAGE is OFF by default and not exercised in CI, this only surfaces downstream. A CI job that builds with the option on would catch this class of breakage.

olantwin:
ROOT 6.40 moved RRawPtrWriteEntry from ROOT::Experimental::Detail to ROOT::Detail. Select the namespace with a ROOT version check so the RNTuple storage backend (FORM_USE_RNTUPLE_STORAGE=ON) compiles against both ROOT 6.38 and 6.40.

Fixes #712.

- **Code**
  - Added ROOT version-aware handling for `RRawPtrWriteEntry`.
  - Uses `ROOT::Detail` with ROOT 6.40+ and `ROOT::Experimental::Detail` for ROOT 6.38.
  - Updated the `m_entry` member to use the compatibility alias.
- **Compatibility**
  - Enables the RNTuple storage backend to compile with ROOT 6.38 and newer, including ROOT 6.40.

### DIFF
--- a/form/root_storage/root_rntuple_write_container.hpp
+++ b/form/root_storage/root_rntuple_write_container.hpp
@@ -5,6 +5,8 @@
 
 #include "storage/storage_write_association.hpp"
 
+#include "RVersion.h"
+
 #include <memory>
 #include <string>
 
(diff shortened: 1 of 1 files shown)


gemmeren (commented):
Thanks Oliver, LGTM, but @aolivier23 should have a look.

aolivier23 (approved):
Looks good to me.  Thank you for pushing this upstream @olantwin @gemmeren

olantwin:
Thanks for the reviews and for accepting the fix!

PR #771, opened 2026-07-27
clang-tidy: resolve `modernize-avoid-c-style-cast`

### LINKED ISSUE #780: Clarify pylong_as_bool invalid-input return value
## Summary

Evaluate the invalid-input return value in `pylong_as_bool` in `plugins/python/src/modulewrap.cpp`.

The function sets a Python `ValueError` and currently returns `static_cast<bool>(-1)`. In C++, that conversion yields `true`, which makes the returned value look like a successful boolean result if it is observed independently of the pending Python exception.

## Required changes

Define and document the intended error-path contract for `pylong_as_bool`, then update the implementation accordingly. Consider whether the return should be simply `true`, `false`, or whether the function/API should be structured so that an error cannot be represented by a boolean return value.

## Rationale

PR #771 replaced a C-style cast with `static_cast<bool>(-1)` to satisfy `modernize-avoid-c-style-cast`. The modernization preserves existing behavior, but the resulting expression exposes an unclear sentinel/error convention that should be reviewed separately.

## Affected area

- `plugins/python/src/modulewrap.cpp`
- `pylong_as_bool(PyObject*)`
- Python-to-C++ boolean conversion through the `BASIC_CONVERTER` machinery

## Backlinks

- PR #771: https://github.com/Framework-R-D/phlex/pull/771
- Review discussion: https://github.com/Framework-R-D/phlex/pull/771#discussion_r3667268785

Requested by @knoepfel.

knoepfel:
Resolve `modernize-avoid-c-style-cast` clang-tidy warnings.

- **Code quality:** Replaced C-style casts with explicit `static_cast` expressions across resource usage, Python wrappers, and module argument handling to resolve `modernize-avoid-c-style-cast` clang-tidy warnings.
- **Tests:** Updated product-selection and type-distinction test helpers to use safer, clearer type conversions.
- **Documentation:** Marked the clang-tidy fix as complete and recorded the associated pull request reference.


plugins/python/src/modulewrap.cpp
```diff
@@ -497,9 +497,9 @@ namespace {
     // fail to pass float -> bool; the problem is rounding (0.1 -> 0 -> False)
     if (!(l == 0 || l == 1) || PyFloat_Check(pyobject)) {
       PyErr_SetString(PyExc_ValueError, "boolean value should be bool, or integer 1 or 0");
-      return (bool)-1;
+      return static_cast<bool>(-1);
```
beojan, line 515: Not really part of this PR so I don't know if it should be a blocker, but what is this?

plugins/python/src/modulewrap.cpp
```diff
@@ -520,15 +520,15 @@ namespace {
...
+    return static_cast<long>(-1);
   }
 
   unsigned long pylong_or_int_as_ulong(PyObject* pyobject)
   {
     // convert <pybject> to C++ unsigned long, with bounds checking, allow int -> ulong.
     if (PyFloat_Check(pyobject)) {
       PyErr_SetString(PyExc_TypeError, "can\'t convert float to unsigned long");
-      return (unsigned long)-1;
+      return static_cast<unsigned long>(-1);
```
beojan, line 546: And again.

plugins/python/src/modulewrap.cpp
```diff
@@ -546,10 +546,10 @@ namespace {
       PyErr_Clear();
       long i = PyLong_AS_LONG(pyobject);
       if (0 <= i) {
-        ul = (unsigned long)i;
+        ul = static_cast<unsigned long>(i);
       } else {
         PyErr_SetString(PyExc_ValueError, "can\'t convert negative value to unsigned long");
-        return (unsigned long)-1;
+        return static_cast<unsigned long>(-1);
```
beojan, line 567: Once more

beojan (dismissed):

plugins/python/src/modulewrap.cpp
knoepfel: This yields the same result as `-1ul` or `std::numeric_limits<unsigned long>::max()`.

plugins/python/src/modulewrap.cpp
knoepfel: Same as `-1ul`.

plugins/python/src/modulewrap.cpp
knoepfel: Yeah...this one is particularly icky.  `static_cast<bool>(-1)` resolves to `true`.  Do we want this?  Will have to ask @wlav.

knoepfel: @coderabbit, can you create an issue that tracks this particular change?  Specifically, we should evaulate whether `static_cast<bool>(-1)` should be replaced with simply `true`, or something else.

plugins/python/src/modulewrap.cpp
beojan: Personally I would prefer `numeric_limits<unsigned long>::max()`, but admittedly that's just preference.

plugins/python/src/modulewrap.cpp
wlav: Leave it as-is. The normal path can be true or false, the `(cast)-1` is a Python convention to prevent having to call `PyErr_Occurred()` all the time.

plugins/python/src/modulewrap.cpp
wlav: No, that's not a preference, it'd be wrong: the `-1` is conventional. Yes, it'd resolve to the same value but the code would no longer reflect the intend and the receiving end would be left to guess as to what they should compare to.

plugins/python/src/modulewrap.cpp
knoepfel: @wlav, which of these two options do you prefer:

1. Keep `(bool)-1` but decorate it with a `// NOLINT` comment to suppress the clang-tidy warning
2. Keep `static_cast<bool>(-1)` as seen in this PR without a `// NOLINT` comment

wlav: I prefer 2.

In the end, it's only important that the receiving end does the same thing and this is a file-local function, so we have that under control. Ie. either change them _all_ or change none. The overall optimization for `bool` is going to be very minimal; just need to prevent false negatives.

knoepfel: Got it.  We'll keep 2 then.

docs/dev/clang-tidy-fixes-2026-04.md
```diff
@@ -101,7 +101,8 @@
 - [x] [modernize-avoid-c-arrays](https://clang.llvm.org/extra/clang-tidy/checks/modernize/avoid-c-arrays.html) (17)
   - _a.k.a._ [cppcoreguidelines-avoid-c-arrays](https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/avoid-c-arrays.html)
   - [PR #745](https://github.com/Framework-R-D/phlex/pull/745)
-- [ ] [modernize-avoid-c-style-cast](https://clang.llvm.org/extra/clang-tidy/checks/modernize/avoid-c-style-cast.html) (82)
+- [x] [modernize-avoid-c-style-cast](https://clang.llvm.org/extra/clang-tidy/checks/modernize/avoid-c-style-cast.html) (82)
+  - [PR #771](https://github.com/Framework-R-D/phlex/pull/771)
```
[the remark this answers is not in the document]
knoepfel: I believe the remaining C-style casts have been addressed.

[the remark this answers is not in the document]
knoepfel: I've now addressed that cast.

beojan (approved):

PR #777, opened 2026-07-28
FORM Unify TTree Testing

aolivier23:
This PR fixes the last FORM test that works differently for TTree and RNTuple.  The TTree backend used to silently allow a call to `commit()` when `fill()` had never been called.  This probably causes a corrupted TTree on file.  Now, FORM's TTree backend will throw an exception in `commit()` when the number of entries in its branch (number of times it has been `fill()`ed) is not greater than the number of times `commit()` has been called.


aolivier23:
@phlexbot format

gemmeren (dismissed):

gemmeren (approved):

wwuoneway (approved):
LGTM — thanks.

PR #784, opened 2026-07-29
clang-tidy: resolve `modernize-use-equals-default`

knoepfel:
Resolve `modernize-use-equals-default` clang-tidy warnings.


aolivier23 (approved):
Looks good to me.

PR #796, opened 2026-07-31
clang-tidy: resolve `performance-unnecessary-value-param`

knoepfel:
Resolve `performance-unnecessary-value-param` clang-tidy warnings.

- **Code quality**
  - Resolved `performance-unnecessary-value-param` clang-tidy warnings.
  - Replaced unnecessary by-value parameters with `const&` or `std::string_view`.
  - Added intentional-move suppressions and updated allowed clang-tidy types.
  - Moved owned objects where required to preserve transfer semantics.

- **API and framework**
  - Updated graph, glue, registration, routing, storage, and plugin APIs.
  - Made applicable proxy creation and registration methods `const`.
  - Changed `flush_gate::roll_up_child` to accept counts by reference.
  - Updated Python wrappers to use const module and source pointers.

- **Correctness**
  - Added assertions for committed flush counts before roll-up.
  - Updated `data_cell_cursor` to use non-owning pointers consistently.
  - Preserved validation, routing, registration, and error behavior.

- **Tests and documentation**
  - Updated flush-gate tests for direct `data_cell_counts` values.
  - Recorded completion of the clang-tidy warning fix in `docs/dev/clang-tidy-fixes-2026-04.md`.


plugins/python/src/modulewrap.cpp
```diff
@@ -61,32 +61,32 @@ using phlex::product_selector;
...
 // clang-format on
 
-PyObject* phlex::experimental::wrap_source(phlex_source_t& source)
+PyObject const* phlex::experimental::wrap_source(phlex_source_t const& source)
 {
   py_phlex_source* pysrc = PyObject_New(py_phlex_source, &PhlexSource_Type);
   pysrc->ph_source = &source;
 
-  return reinterpret_cast<PyObject*>(pysrc);
+  return reinterpret_cast<PyObject const*>(pysrc);
```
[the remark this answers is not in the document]
knoepfel: The latest commit should resolve this

phlex/detail/plugin_macros.hpp
```diff
@@ -13,10 +13,10 @@
 #define PHLEX_DETAIL_NARGS(...) BOOST_PP_DEC(BOOST_PP_VARIADIC_SIZE(__VA_OPT__(, ) __VA_ARGS__))
 
 #define PHLEX_DETAIL_CREATE_1ARG(token_type, func_name, m)                                         \
-  void func_name(token_type<phlex::detail::void_tag> m, phlex::configuration const&)
+  void func_name(token_type<phlex::detail::void_tag> const& m, phlex::configuration const&)
 
 #define PHLEX_DETAIL_CREATE_2ARGS(token_type, func_name, m, cfg)                                   \
-  void func_name(token_type<phlex::detail::void_tag> m, phlex::configuration const& cfg)
+  void func_name(token_type<phlex::detail::void_tag> const& m, phlex::configuration const& cfg)
```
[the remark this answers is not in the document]
knoepfel: The latest commit resolves this

phlex/core/graph_proxy.hpp
```diff
@@ -47,48 +48,60 @@ namespace phlex::detail {
...
 
     /// @brief Registers a source (used by the framework to create provider nodes)
     template <std::derived_from<source> Source, typename... Args>
-    void add_source(std::string name, Args&&... args)
+    void add_source(std::string_view name, Args&&... args) const
       requires(not is_bound_object<T>);
 
     /// @brief Registers an output node.
-    auto output(std::string name, is_output_like auto f, concurrency c = concurrency::serial);
+    auto output(std::string_view name, is_output_like auto f, concurrency c = concurrency::serial);
```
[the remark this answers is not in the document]
knoepfel: Resolved with the latest commit

phlex/core/make_computational_edges.cpp
```diff
@@ -39,7 +39,7 @@ namespace phlex::detail {
     }
 
     std::pair<index_router::provider_input_ports_t, index_router::head_ports_t>
-    edges_from_explicit_providers(index_router::head_ports_t head_ports,
+    edges_from_explicit_providers(index_router::head_ports_t const& head_ports,
```
beojan, line 42: We could instead take a non-const reference to `head_ports` then `std::swap` `head_ports` and `unconsumed_head_ports` before returning. This would also mean we could return just the `provider_input_ports_t` instead of a pair.

This also applies to `edges_from_implicit_providers`.

knoepfel: @beojan, I think in this case, keeping the `const` there makes it easier to understand what is mutable and what is not.  If this becomes a performance bottleneck, then we can adjust later on.

phlex/core/glue.hpp
```diff
@@ -52,14 +53,19 @@ namespace phlex::detail {
...
+              concurrency c,
+              std::string partition,
+              InitArgs&&... init_args)
     {
       internal::verify_name(name, config_);
       return fold_api{config_,
-                      std::move(name),
-                      algorithm_bits(bound_obj_, std::move(f)),
+                      name,
+                      algorithm_bits(std::move(bound_obj_), std::move(f)),
```
marcpaterno: This could be dangerous. Unless we can be sure that the data member bound_obj_ can never be used again (or that every access is checked so that if the moved from shared_ptr null, the access does not dereference bound_obj_ without checking it first), it may be better to just copy the shared_ptr.

This comment also applies to the several other places where bound_obj_ is moved from.

phlex/core/framework_graph.cpp
```diff
@@ -110,7 +110,7 @@ namespace phlex::detail {
                     tbb::flow::unlimited,
                     [this](ready_flushes_then_emit const& input) -> data_cell_index_ptr {
                       auto&& [ready_flushes, index_to_emit] = input;
```
marcpaterno: Consider changing `auto&&` to the more constrained `auto const&`.

phlex/model/fixed_hierarchy.hpp
```diff
@@ -35,12 +35,10 @@ namespace phlex {
...
     data_cell_index_ptr index_;
-    // Non-owning references to the enclosing hierarchy and driver; data_cell_cursor is a
-    // short-lived view and does not manage their lifetimes.
-    // NOLINTBEGIN(cppcoreguidelines-avoid-const-or-ref-data-members)
-    fixed_hierarchy const& hierarchy_;
-    detail::framework_driver& driver_;
-    // NOLINTEND(cppcoreguidelines-avoid-const-or-ref-data-members)
+    // Non-owning pointers to the enclosing hierarchy and driver; data_cell_cursor is a
+    // short-lived, freely copyable view and does not manage their lifetimes.
+    fixed_hierarchy const* hierarchy_;
```
marcpaterno, line 40: If we already depend on the Guidelines Support Library (gsl, not the other gsl) then consider use of their `gsl::not_null<T>` for these pointers.

knoepfel: After exploring how to do this, adding `gsl::not_null<T>` would introduce some public-facing dependency changes that we need to sort through.  My suggestion now is to create a GitHub issue that tracks this (desirable) change, but to implement it as part of a separate PR.

@coderabbit, can you add a GitHub issue for this?

phlex/core/framework_graph.cpp
knoepfel: Resolved with commit https://github.com/Framework-R-D/phlex/pull/796/commits/02fb07794439a5c10ef93cfba028672a77c7cb8f.

phlex/core/glue.hpp
knoepfel: Addressed with commit https://github.com/Framework-R-D/phlex/pull/796/commits/8ddb7a6f697d31080bd4ddcf5f32cc152682c8df.

knoepfel: See issue https://github.com/Framework-R-D/phlex/issues/825.

marcpaterno (approved):

PR #800, opened 2026-08-03
Fix set construction when deciding to insert repeaters

beojan:
Previously we were constructing a set of two iterators, and therefore unconditionally inserting repeaters.

Ultimately it doesn't look like this saves any appreciable time on the tests so we might as well remove the condition when we make layer names optional.

- **Code**
  - Fixed `collapsed_layers` set construction in `phlex/core/multilayer_join_node.hpp`.
  - The set now contains layer values instead of two iterators.
  - Repeaters are inserted only when the required layers are absent.


knoepfel (approved):
Good catch, @beojan!

PR #801, opened 2026-08-04
chore(deps): add npm release age constraint for @kilocode/cli

greenc-FNAL:
- Set min-release-age=3 to defer installation of npm packages released within
  the last 3 days, reducing risk of installing unstable prereleases
- Add .npmrc config file to persist this setting across builds
- Update Dockerfile to configure npm before installing @kilocode/cli

- **Build system**
  - Add `.npmrc` with `min-release-age=3`.
  - Configure npm with the three-day release-age limit before installing `@kilocode/cli` in `.devcontainer/Dockerfile`.
  - Defer installation of packages released within the previous three days.

- **CI**
  - Update the Markdown Fix and YAML Fix workflows to use Node.js v26.
  - Disable the setup-node package-manager cache to avoid stale npm dependencies.
  - Install `markdownlint-cli2` with a multiline shell command in the Markdown Fix workflow.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -103,6 +103,7 @@ INSTALL_DEV_UTILS
 
 RUN bash <<'INSTALL_NPM_PACKAGES'
 set -euo pipefail
+npm update -g npm@^11.10.0
 npm config set allow-scripts=@kilocode/cli --location=user
 npm config set min-release-age=3 --location=user
 npm install -g @kilocode/cli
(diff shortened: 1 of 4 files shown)


greenc-FNAL:
@coderabbitai review

knoepfel (commented):
How is the top-level `.npmrc` file used?  It would be nice if it didn't have to go in the top-level directory of the repository, which is already cluttered with many files.

greenc-FNAL:
> How is the top-level `.npmrc` file used? It would be nice if it didn't have to go in the top-level directory of the repository, which is already cluttered with many files.

Good call: this is an AI mistake. The "project-level" `.npmrc` file is only valid where `npm_modules` is a sibling, and is not read at all for global (`-g`) installs. 

I will set it at the user level instead of relying on a file (_cf_ `.devcontainer/Dockerfile`).

knoepfel (approved):

PR #805, opened 2026-08-05
clang-tidy: resolve `readability-use-anyofallof`

knoepfel:
Resolve `readability-use-anyofallof` clang-tidy warnings.

### DIFF
--- a/plugins/python/src/modulewrap.cpp
+++ b/plugins/python/src/modulewrap.cpp
@@ -897,69 +897,6 @@ static std::optional<std::string_view> collection_dtype(std::string const& type_
   return std::string_view{type_name}.substr(pos);
 }
 
-static bool insert_input_converter(py_phlex_module* mod,
-                                   std::string const& cname, // TODO: shared_ptr<PyObject>
-                                   size_t i,
-                                   product_selector const& inp_pq,
-                                   std::string const& inp_type,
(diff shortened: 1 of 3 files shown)


plugins/python/src/modulewrap.cpp
```diff
@@ -904,61 +967,11 @@ static bool insert_input_converters(py_phlex_module* mod,
...
-      }
-    } else {
-      PyErr_Format(PyExc_TypeError, "unsupported input type \"%s\"", inp_type.c_str());
-      return false;
-    }
-  }
-
-  return true;
+  auto const converters = std::views::zip(std::views::iota(size_t{}), input_selectors, input_types);
+  return std::ranges::all_of(converters, [mod, &cname, ispy, nc](auto const& converter) {
```
beojan: I'm not sure `all_of` is appropriate here. We have a loop where the side effects are what's important.

marcpaterno: I do not think the standard promised that `all_of` short circuits, which is what we want. So I agree we should not use `all_of` here.

knoepfel: The range-for loop has been restored, with its short-circuiting behavior.

marcpaterno (approved):

PR #813, opened 2026-08-06
refactor: clean up sendability naming and flush-gate count typing

### LINKED ISSUE #812: Rename move_constructible_only to clarify sendability semantics
## Summary

Rename the `phlex::experimental::move_constructible_only` concept to a name that describes its role as the fallback sendability path for move-constructible types.

## Rationale

`move_constructible_only` can be read as saying that move construction is the only way to create the type. The concept actually selects types that are move-constructible and do not provide a `send(...)` overload. A name such as `sendable_by_move` would make this intent clearer.

## Affected area

- `phlex/core/fold/send.hpp`
- `phlex::experimental::move_constructible_only`
- The constrained `sendable_type_impl` specialization that uses this concept

## Acceptance criteria

- Replace `move_constructible_only` with a clearer name that describes sendability through move construction.
- Update all references in `phlex/core/fold/send.hpp`.
- Keep the existing overload-precedence behavior: use `send(...)` when available; otherwise use the move-constructible type.

## Backlinks

- Pull request: https://github.com/Framework-R-D/phlex/pull/540
- Review comment: https://github.com/Framework-R-D/phlex/pull/540#discussion_r3730370630
- Requested by: @knoepfel

knoepfel:
- **Code**
  - Rename `move_constructible_only` to `sendable_by_move`.
  - Store `flush_gate::committed_counts_` by value instead of through a shared pointer.
  - Update `committed_counts()` and `roll_up_child()` to use references.
  - Remove pointer dereferencing, null handling, and unused includes.

- **Tests**
  - Update flush-gate tests to construct `data_cell_counts` by value.
  - Preserve spill-count accumulation and child-to-parent propagation coverage.

Replaces storing a committed-counts data member as a shared pointer by storing it by value instead.
Before PR https://github.com/Framework-R-D/phlex/pull/540 was merged, the committed-counts data member needed to be sent to downstream
nodes.  The type of the member was not copyable, so a shared pointer to that type was used instead.
This has been fixed with this PR.

### DIFF
--- a/phlex/model/flush_gate.hpp
+++ b/phlex/model/flush_gate.hpp
@@ -40,6 +40,7 @@
 #include <functional>
 #include <memory>
 #include <mutex>
+#include <utility>
 
 namespace phlex::detail {
 
@@ -57,7 +58,7 @@ namespace phlex::detail {
     std::size_t expected_total_count() const;
     std::size_t committed_total_count() const;
     std::size_t committed_count_for_layer(data_cell_index::hash_type layer_hash) const;
-    data_cell_counts_const_ptr committed_counts() const { return committed_counts_; }
+    data_cell_counts const& committed_counts() const { return committed_counts_; }
 
     // Merges an expected child count into the accumulated expected counts.  Each call
     // represents one flush message arriving (e.g. one unfold completing for this index).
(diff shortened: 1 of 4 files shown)


phlex/model/flush_gate.hpp
```diff
@@ -84,11 +85,7 @@ namespace phlex::detail {
 
     data_cell_index_ptr const index_;
     std::once_flag commit_once_;
-    // FIXME: We express committed_counts_ as a shared pointer so that we can copy the committed
-    //        counts (this is done for determining the flush values for folds).  Once the fold
-    //        flushes are incorporated as part of the multi-layer join node infrastructure, it
-    //        should be possible for committed_counts_ to no longer be a pointer, but a value.
```
pcanal, line 90: It might be helpful to add to the commit log a summary of why the `shared_ptr` is no longer needed (eg. could include the PR number(s) that render this moot).

pcanal (approved):
LGTM (although the commit log commit enhance with more detail on the 'why' of the changes).

phlex/model/flush_gate.hpp
knoepfel: I will add the explanation to the commit message.

PR #816, opened 2026-08-07
Form redesign backend technology

wwuoneway:
**Summary:**
Redesign of how a backend's storage technology is represented:
- Previously the technology was a single int that encoded two levels through arithmetic (major*256 + minor), packed and unpacked via Combine / GetMajor / GetMinor, and re-parsed from strings independently in every module, source, and test. 
- This PR replaces it with a typed value in the foundational (core/) layer.

**What changed:**
- Typed `technology::Id` replaces the packed `int`. `Major` is an `enum`; `minor` selects the variant within a major (e.g., TTree vs RNTuple within ROOT).
- Defaulted operator`<=>` gives exact (major, minor) ordering, so Id is usable directly as a std::map key and for backend dispatch.
- Single `from_string` / `to_string` consolidates the three duplicated parse tables that had drifted across the codebase.
- Lives in `core/` (not `form/`), so the core carriers `Placement` and `Token` don't depend upward on `form/`.

**Compatibility/Consistence:**
- naming style (e.g., `form::technology::Id`) will need a further discussion. A scan across the form/ and test/form/ folders for consistence check will be needed once we decide. This could be split into a separate PR.

- **Core API**
  - Added typed `technology::Id` with `Major`, `minor`, ordering, and predefined ROOT/HDF5 identifiers.
  - Added centralized `from_string` and `to_string` conversion.
  - Moved technology definitions from `form/` to `core/`.

- **FORM integration**
  - Updated `Placement`, `Token`, configuration, storage, and factory APIs to use `technology::Id`.
  - Replaced packed integer comparisons and local parsing tables with typed IDs and centralized conversion.
  - Preserved backend lookup and ROOT TTree/RNTuple dispatch behavior.

- **Tests**
  - Updated FORM tests and utilities to use typed technology IDs.
  - Updated technology string handling and generated test filenames.


form/core/technology.hpp
```diff
@@ -0,0 +1,66 @@
...
+    int minor{0};
+
+    // Exact ordering over (major, minor): lets an Id be a std::map key and drives backend dispatch
+    constexpr auto operator<=>(Id const&) const = default;
+  };
+
+  // Backends: valid (major, minor) pairs, stable numeric values as a future Token may persist them
+  inline constexpr Id ROOT_TTREE{Major::root, 1};
+  inline constexpr Id ROOT_RNTUPLE{Major::root, 2};
+  inline constexpr Id HDF5{Major::hdf5, 1};
```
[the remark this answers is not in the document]
wwuoneway: As noted in the PR description, the naming and namespace for this API still need broader discussion within form. form/ does not currently follow the phlex/ core conventions consistently, so changing only this PR would make that inconsistency more visible rather than solve it fully. 
I’m happy to address this in a follow-up PR with a broader consistency pass across form/ and test/form/ once we align on the convention.

gemmeren (dismissed):

form/core/technology.hpp
```diff
@@ -0,0 +1,66 @@
...
+  inline Id from_string(std::string_view name)
+  {
+    if (name == "ROOT_TTREE") {
+      return ROOT_TTREE;
+    }
+    if (name == "ROOT_RNTUPLE") {
+      return ROOT_RNTUPLE;
+    }
+    if (name == "HDF5") {
+      return HDF5;
```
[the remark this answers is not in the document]
wwuoneway: Done, now throw "Technology 'HDF5' is recognized but not yet implemented"

form/core/technology.hpp
```diff
@@ -0,0 +1,67 @@
...
+  // Minor variant within a Major (e.g. TTree vs RNTuple within ROOT)
+  struct Id {
+    Major major{Major::unknown};
+    int minor{0};
+
+    // Exact ordering over (major, minor): lets an Id be a std::map key and drives backend dispatch
+    constexpr auto operator<=>(Id const&) const = default;
+  };
+
+  // Backends: valid (major, minor) pairs, stable numeric values as a future Token may persist them
```
aolivier23, line 29: Why did you choose to make major an enum class but minor just an integer?  Could they both be enum classes?

form/util/factories.hpp
```diff
@@ -25,91 +25,88 @@
...
 #ifdef USE_ROOT_STORAGE
       return std::make_shared<ROOT_TFileImp>(name, mode);
 #endif
-    } else if (form::technology::GetMajor(tech) == form::technology::HDF5_MAJOR) {
-      // Handle HDF5 file creation when implemented
-      // return std::make_shared<HDF5_FileImp>(name, mode);
+    } else if (tech.major == form::technology::Major::hdf5) {
+      throw std::runtime_error("FORM: HDF5 storage is recognized but not yet implemented");
     }
     return std::make_shared<Storage_File>(name, mode);
```
aolivier23: We can still get here if someone passes a technology that FORM doesn't know about, like Apache Arrow.  Can we throw an exception at the end of the whole function instead?

form/util/factories.hpp
```diff
@@ -25,91 +25,88 @@
...
 
-  inline std::shared_ptr<IStorage_Write_Container> createWriteContainer(int tech,
+  inline std::shared_ptr<IStorage_Write_Container> createWriteContainer(form::technology::Id tech,
                                                                         std::string const& name)
   {
-    // Use the helper functions from Technology namespace for consistency
-    if (form::technology::GetMajor(tech) == form::technology::ROOT_MAJOR) {
-      if (form::technology::GetMinor(tech) == form::technology::ROOT_TTREE_MINOR) {
+    if (tech.major == form::technology::Major::root) {
+#if defined(USE_ROOT_STORAGE) || defined(USE_RNTUPLE_STORAGE)
```
aolivier23: I'm not sure RNTuple works with USE_ROOT_STORAGE turned off.  That was not my intent when I installed the USE_RNTUPLE_STORAGE flag.  What's the reason we should change that?

aolivier23 (commented):
Mostly looks good.  Thank you for this improvement!  There are two comments I left about things I want to discuss:
- if we want HDF5 container creation to throw for now, do we need to also protect against unknown technologies?
- Looks like we're changing the meaning of USE_RNTUPLE_STORAGE in one place but not the others.  Let's discuss that please.

form/core/technology.hpp
wwuoneway: The asymmetry is intentional: `major` is a closed, globally meaningful set of storage families, so an `enum` class fits well. `minor` is only meaningful within a given major—it’s a family-local variant ID, so the same integer can be reused across different majors. Technically, both could be `enum` classes, but that wouldn’t help, and neither alternative is an improvement.

form/util/factories.hpp
wwuoneway: Good catch — done. Each factory now switches on tech.major with an explicit case per known Major, and a throw std::runtime_error(...) sits after the switch as the catch-all. So an unrecognized technology like Apache Arrow now fails loudly at the end of the function instead of falling through to the data-discarding generic backend.

form/util/factories.hpp
wwuoneway: Agreed — RNTuple should not be enabled without ROOT, and that is now enforced explicitly. The runtime path was already effectively guarded (now in `form/storage/factories.cpp`), the RNTuple backend is only compiled within the ROOT-enabled code path, so ROOT_RNTUPLE cannot be used when ROOT is off and now throws instead of degrading silently. See `form/CMakeLists.txt` for the added configure-time error to make the dependency explicit. So there’s no intent to decouple the two flags — this change just makes the original intent explicit. Later, I’ll also check across the package to refine the `USE_ROOT_STORAGE` / `USE_RNTUPLE_STORAGE` flag handling for consistency.

form/storage/factories.cpp
```diff
@@ -0,0 +1,136 @@
...
+namespace form::detail::experimental {
+
+  using Major = form::technology::Major;
+
+  std::shared_ptr<IStorage_File> createFile(form::technology::Id tech,
+                                            std::string const& name,
+                                            char mode)
+  {
+    switch (tech.major) {
+    case Major::unknown:
```
gemmeren: `:unknown:` seems wrong here and would make me believe we should throw... How about `:generic:` or so?

gemmeren (dismissed):

form/core/technology.hpp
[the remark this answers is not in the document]
wwuoneway: No need to open a separate issue for now. I'll address this in a follow-up PR as part of the broader consistency pass mentioned in the PR description.

aolivier23 (dismissed):
LGTM.  Thanks for the discussion.

form/storage/factories.cpp
wwuoneway: good point. I agree that `unknown` is not ideal here. Since this PR has already passed all checks and received approval, and this is only a minor naming detail, I'd prefer to leave it for a follow-up. Once we clarify the user configuration and default/unspecified technology semantics more broadly, we can address this consistently.

pcanal: If addressing this comment after the PR, please either open the follow PR or a reminder issue and link it here so that we can keep track of the outstanding conversation.  Thanks.

form/core/technology.hpp
```diff
@@ -0,0 +1,67 @@
...
+#include <stdexcept>
+#include <string>
+#include <string_view>
+
+/* A storage technology, identified by a (major, minor) pair */
+
+namespace form::technology {
+
+  // Major storage type (ROOT, HDF5, ...)
+  enum class Major : int {
```
pcanal: Is the size specification meaningful?  (`int` is already the default). Should it be shorter?

form/storage/factories.cpp
```diff
@@ -0,0 +1,136 @@
...
+      // No technology specified: generic storage.
+      return std::make_shared<Storage_Read_Container>(name);
+    case Major::root:
+#ifdef USE_ROOT_STORAGE
+      if (tech == form::technology::ROOT_TTREE) {
+        return std::make_shared<ROOT_TBranch_Read_ContainerImp>(name);
+      } else if (tech == form::technology::ROOT_RNTUPLE) {
+#ifdef USE_RNTUPLE_STORAGE
```
pcanal, line 90: What is the use case from builds without `RNTuple` supports?

pcanal (dismissed):
LFTM

form/storage/factories.cpp
wwuoneway: thanks, will do.

form/core/technology.hpp
wwuoneway: Good point. I agree that the explicit `: int` could be omitted. Given that the PR is already approved, I think it's best to leave this minor cleanup for a follow-up consistency pass.

pcanal: > Given that the PR is already approved, 

I do not agree with this philosophy per se.  For example my approval is (per the actual policy) conditional to the resolution of the conversations (to the satisfaction of the reviewer and the author).  As stated elsewhere if a conversation require large enough change that it deserve to open an issue then that is an acceptable solution.  If a conservation only requires minor changes (for example in this case), then they should be done within the ongoing PR rather than incurring the cost of opening an issue and creating a new PR.

form/storage/factories.cpp
gemmeren: Hi Philippe, this is for development only and hopefully will go away with Phlex 0.4... Until then we wanted to develop RNTuple backend in a more experimental fashion. After that we should collapse the build flags into a single ROOT dependency and protect RNTuple via ROOT_Version as needed.

form/core/technology.hpp
gemmeren: @wwuoneway , how much effort would it be to address minor comments in new commits to _this_ PR ? I believe if we all agree on a change and it isn't too complex, then re-running a few CI's and re-approving the PR should not stand in the way of making the code change better.

form/storage/factories.cpp
gemmeren: This too is a rather minor change and the interpretation of a technology being `:generic:` is significantly different from it being `:unknown:` ... Please consider renaming as part of *this* PR.

form/core/technology.hpp
wwuoneway: Done — `int` is removed.

form/storage/factories.cpp
wwuoneway: Done — renamed `Major::unknown` → `Major::generic`.

gemmeren (dismissed):

form/storage/factories.cpp
pcanal: Fair enough.  Thanks.

form/storage/factories.cpp
```diff
@@ -0,0 +1,136 @@
...
+        return std::make_shared<ROOT_TTree_Write_ContainerImp>(name);
+      } else if (tech == form::technology::ROOT_RNTUPLE) {
+#ifdef USE_RNTUPLE_STORAGE
+        return std::make_shared<ROOT_RNTuple_Write_ContainerImp>(name);
+#else
+        throw std::runtime_error("FORM: ROOT RNTUPLE support is not compiled into this build");
+#endif
+      }
+      // Recognized ROOT major, but an unsupported subtype/minor.
+      throw std::runtime_error("FORM: requested ROOT write-association backend is not available");
```
pcanal, line 68: According to codecov: see [factories.cpp]( https://app.codecov.io/gh/Framework-R-D/phlex/pull/816?src=pr&el=tree&filepath=form%2Fstorage%2Ffactories.cpp&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Framework-R-D#41e446f1c8dd33ccb90fc38489d3c163-R126.).

This line (68) and line 97 are not tested.   Can you add test(s) that check that if the user passes an unsupported value for the technology then we get the expected exception?

Thanks.

form/storage/factories.cpp
```diff
@@ -0,0 +1,136 @@
...
+#ifdef USE_RNTUPLE_STORAGE
+        return std::make_shared<ROOT_RField_Read_ContainerImp>(name);
+#else
+        throw std::runtime_error("FORM: ROOT RNTUPLE support is not compiled into this build");
+#endif
+      }
+      // Recognized ROOT major, but an unsupported subtype/minor.
+      throw std::runtime_error("FORM: requested ROOT read-container backend is not available");
+#else
+      throw std::runtime_error("FORM: ROOT support is not compiled into this build");
```
pcanal, line 99: See https://github.com/Framework-R-D/phlex/pull/816/changes#r3780620017

aolivier23 (dismissed):
LGTM.  I agree with Phillipe about the tests, but they're not a show-stopper for me.

aolivier23 (approved):
Thank you for the new test!  LGTM.

gemmeren (approved):

form/storage/factories.cpp
wwuoneway: Done — thanks.

form/storage/factories.cpp
wwuoneway: Done — thanks.

PR #817, opened 2026-08-10
Make layer specification optional

### LINKED ISSUE #875: Layer field optional in `product_selector` under some circumstances
The following behavior is intended for the data layer:

1. Explicit providers must specify a single data layer.
2. Each automatically generated implicit provider is configured to read from a single data layer.
3. An omitted layer in a `product_selector` is a wildcard during matching, but provider resolution must select a single provider and concrete data layer. Ambiguous explicit-provider matches will be rejected for now.
4. Optional layer specifications are supported only by single-input computational operators. Multi-input operators must specify layers for all inputs.
5. Fold and unfold selectors continue to require layers because their flush machinery depends on them.

beojan:
Layer specifications are now optional in `product_selector`s. As discussed with Kyle and Marc, this is currently limited to single input functions. Folds are always considered multi-input (they have an implicit accumulator input).

In addition, this is not currently enabled for unfolds because their input layers are used in the flush machinery.

_AI summary below_
___

- **Core behavior**
  - Allows omitted `product_selector` layers for single-input functions.
  - Resolves omitted layers from explicit or implicit providers.
  - Treats omitted query layers as wildcards during selector matching.
  - Keeps folds as multi-input functions and requires their layers.
  - Rejects unfold inputs without layers because flush machinery uses those layers.
  - Improves diagnostics for missing and invalid layer names.

- **Implementation**
  - Adds optional layer storage and validation in `product_selector`.
  - Updates provider wiring and implicit-provider reuse.
  - Uses wildcard identifiers for unresolved provider input layers.
  - Updates `products_consumer` to enforce layers for multi-input consumers.
  - Adds required compiler-specific linking for GNU toolchains.

- **Tests**
  - Covers layerless selectors for computational nodes, predicates, observers, transformed products, and explicit or implicit providers.
  - Verifies shared implicit-provider inputs and execution counts.
  - Verifies the expected diagnostic for layerless unfold inputs.

### DIFF
--- a/phlex/core/CMakeLists.txt
+++ b/phlex/core/CMakeLists.txt
@@ -1,13 +1,3 @@
-if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
-  if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 14.1)
-    set(_std_stacktrace_library stdc++exp)
-  else()
-    set(_std_stacktrace_library stdc++_libbacktrace)
-  endif()
-elseif(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
-  set(_std_stacktrace_library stdc++exp)
-endif()
-
 cet_make_library(
   LIBRARY_NAME
   phlex_core
@@ -44,7 +34,6 @@ cet_make_library(
   phlex::metaprogramming
   phlex::model
   phlex::utilities
-  ${_std_stacktrace_library}
   PRIVATE
   Boost::json
   spdlog::spdlog
@@ -106,7 +95,6 @@ phlex_make_internal_library(
   phlex_core
   LIBRARIES
   PUBLIC TBB::tbb phlex::metaprogramming phlex_model_internal phlex_utilities_internal
-         ${_std_stacktrace_library}
   PRIVATE Boost::json spdlog::spdlog
 )
 add_library(phlex::core_internal ALIAS phlex_core_internal)
--- a/phlex/core/CMakeLists.txt
+++ b/phlex/core/CMakeLists.txt
@@ -38,6 +38,15 @@ cet_make_library(
   Boost::json
   spdlog::spdlog
 )
+
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "GNU")
+  if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 14.1)
+    target_link_libraries(phlex_core PUBLIC -lstdc++exp)
+  else()
+    target_link_libraries(phlex_core PUBLIC -lstdc++_libbacktrace)
+  endif()
+endif()
+
 install(
   FILES
     concepts.hpp
--- a/phlex/core/CMakeLists.txt
+++ b/phlex/core/CMakeLists.txt
@@ -1,3 +1,13 @@
+if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+  if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 14.1)
+    set(_std_stacktrace_library stdc++exp)
+  else()
+    set(_std_stacktrace_library stdc++_libbacktrace)
+  endif()
+elseif(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+  set(_std_stacktrace_library stdc++exp)
+endif()
+
 cet_make_library(
   LIBRARY_NAME
   phlex_core
@@ -34,18 +44,11 @@ cet_make_library(
   phlex::metaprogramming
   phlex::model
   phlex::utilities
+  ${_std_stacktrace_library}
   PRIVATE
   Boost::json
   spdlog::spdlog
 )
-
-if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "GNU")
-  if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 14.1)
-    target_link_libraries(phlex_core PUBLIC -lstdc++exp)
-  else()
-    target_link_libraries(phlex_core PUBLIC -lstdc++_libbacktrace)
-  endif()
-endif()
 
 install(
   FILES
@@ -103,17 +106,10 @@ phlex_make_internal_library(
   phlex_core
   LIBRARIES
   PUBLIC TBB::tbb phlex::metaprogramming phlex_model_internal phlex_utilities_internal
+         ${_std_stacktrace_library}
   PRIVATE Boost::json spdlog::spdlog
 )
 add_library(phlex::core_internal ALIAS phlex_core_internal)
-
-if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "GNU")
-  if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 14.1)
-    target_link_libraries(phlex_core_internal PUBLIC -lstdc++exp)
-  else()
-    target_link_libraries(phlex_core_internal PUBLIC -lstdc++_libbacktrace)
-  endif()
-endif()
 
 # Interface library
 cet_make_library(
--- a/phlex/core/CMakeLists.txt
+++ b/phlex/core/CMakeLists.txt
@@ -107,6 +107,14 @@ phlex_make_internal_library(
 )
 add_library(phlex::core_internal ALIAS phlex_core_internal)
 
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "GNU")
+  if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 14.1)
+    target_link_libraries(phlex_core_internal PUBLIC -lstdc++exp)
+  else()
+    target_link_libraries(phlex_core_internal PUBLIC -lstdc++_libbacktrace)
+  endif()
+endif()
+
 # Interface library
 cet_make_library(
   LIBRARY_NAME
--- a/phlex/core/framework_graph.cpp
+++ b/phlex/core/framework_graph.cpp
@@ -66,10 +66,7 @@ namespace phlex::detail {
         phlex::experimental::identifier const child_layer{n->child_layer()};
         std::set<phlex::experimental::identifier> input_layers_for_unfold;
         for (auto const& input : n->input()) {
-          if (!input.layer) {
-            throw std::runtime_error(fmt::format(
-              "{} used by unfold {} has no layer defined", input, n->name().to_string()));
-          }
+          // Existence of layer already validated
           auto const& input_layer =
             static_cast<phlex::experimental::identifier const&>(input.layer);
           if (not input_layers_for_unfold.insert(input_layer).second) {
--- a/phlex/core/framework_graph.cpp
+++ b/phlex/core/framework_graph.cpp
@@ -66,11 +66,12 @@ namespace phlex::detail {
         phlex::experimental::identifier const child_layer{n->child_layer()};
         std::set<phlex::experimental::identifier> input_layers_for_unfold;
         for (auto const& input : n->input()) {
+          if (!input.layer) {
+            throw std::runtime_error(fmt::format(
+              "{} used by unfold {} has no layer defined", input, n->name().to_string()));
+          }
           auto const& input_layer =
             static_cast<phlex::experimental::identifier const&>(input.layer);
-          if (input_layer.empty()) {
-            continue;
-          }
           if (not input_layers_for_unfold.insert(input_layer).second) {
             continue;
           }
--- a/phlex/core/index_router.cpp
+++ b/phlex/core/index_router.cpp
@@ -108,6 +108,7 @@ namespace phlex::detail {
                               fold_partition_ports_t fold_partition_ports,
                               std::map<std::string, named_index_ports> const& multilayer_join_ports)
   {
+    using namespace phlex::experimental::literals;
     // We must have at least one provider port, or there can be no data to process.
     assert(!provider_input_ports.empty());
 
@@ -196,7 +197,7 @@ namespace phlex::detail {
                                               provider_input_ports_t provider_input_ports)
   {
     for (auto& [input_product, provider_port] : provider_input_ports | std::views::values) {
-      auto [it, _] = index_set_nodes_.emplace(input_product.layer,
+      auto [it, _] = index_set_nodes_.emplace(input_product.layer ? *input_product.layer : "*"_id,
                                               std::make_shared<internal::index_set_node>(g));
       make_edge(*it->second, *provider_port);
     }
--- a/phlex/core/multilayer_join_node.hpp
+++ b/phlex/core/multilayer_join_node.hpp
@@ -83,7 +83,7 @@ namespace phlex::detail {
       std::set collapsed_layers(layers_.begin(), layers_.end());
 
       // Add repeaters only if the inputs span more than one distinct layer.
-      if (collapsed_layers.size() > 1 || collapsed_layers.contains("*"_id)) {
+      if (collapsed_layers.size() > 1) {
         repeaters_.reserve(NInputs);
         for (auto const& layer : layers_) {
           repeaters_.push_back(std::make_unique<internal::repeater_node>(g, name_, layer));
--- a/phlex/core/multilayer_join_node.hpp
+++ b/phlex/core/multilayer_join_node.hpp
@@ -76,15 +76,14 @@ namespace phlex::detail {
       name_{std::move(node_name)},
       layers_{std::move(layer_names)}
     {
-      assert(NInputs == layers_.size());
-
+      using namespace experimental::literals;
       // Collapse to the set of distinct layer names.  More than one distinct layer means
       // at least one input crosses a layer boundary and therefore every input stream
       // needs a repeater_node.
       std::set collapsed_layers(layers_.begin(), layers_.end());
 
       // Add repeaters only if the inputs span more than one distinct layer.
-      if (collapsed_layers.size() > 1) {
+      if (collapsed_layers.size() > 1 || collapsed_layers.contains("*"_id)) {
         repeaters_.reserve(NInputs);
         for (auto const& layer : layers_) {
           repeaters_.push_back(std::make_unique<internal::repeater_node>(g, name_, layer));
--- a/phlex/core/multilayer_join_node.hpp
+++ b/phlex/core/multilayer_join_node.hpp
@@ -76,7 +76,7 @@ namespace phlex::detail {
       name_{std::move(node_name)},
       layers_{std::move(layer_names)}
     {
-      using namespace experimental::literals;
+      assert(NInputs == layers_.size());
       // Collapse to the set of distinct layer names.  More than one distinct layer means
       // at least one input crosses a layer boundary and therefore every input stream
       // needs a repeater_node.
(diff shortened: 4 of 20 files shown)


phlex/core/multilayer_join_node.hpp
```diff
@@ -76,8 +76,7 @@ namespace phlex::detail {
       name_{std::move(node_name)},
       layers_{std::move(layer_names)}
     {
-      assert(NInputs == layers_.size());
-
+      using namespace experimental::literals;
```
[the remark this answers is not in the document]
beojan: Was this part of this PR?

phlex/core/framework_graph.cpp
```diff
@@ -66,11 +66,12 @@ namespace phlex::detail {
         phlex::experimental::identifier const child_layer{n->child_layer()};
         std::set<phlex::experimental::identifier> input_layers_for_unfold;
         for (auto const& input : n->input()) {
+          if (!input.layer) {
+            throw std::runtime_error(fmt::format(
+              "{} used by unfold {} has no layer defined", input, n->name().to_string()));
+          }
```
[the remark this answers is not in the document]
beojan: Was this part of this PR?

phlex/core/products_consumer.hpp
```diff
@@ -21,7 +21,8 @@ namespace phlex::detail {
   public:
     products_consumer(phlex::experimental::algorithm_name name,
                       std::vector<std::string> predicates,
-                      product_selectors input_products);
+                      product_selectors input_products,
+                      bool always_require_layers = false);
```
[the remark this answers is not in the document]
beojan: Addressed

phlex/core/framework_graph.cpp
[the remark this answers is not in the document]
beojan: Can you explain the issue more fully, please?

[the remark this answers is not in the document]
beojan: This should be resolved

beojan:
@codex review

phlex/core/product_selector.hpp
```diff
@@ -42,42 +44,58 @@ namespace phlex {
...
       // NOLINTNEXTLINE(google-explicit-constructor) - Implicit conversion is intentional
-      operator T const&() const noexcept { return content_; }
-      explicit operator std::string_view() const noexcept { return std::string_view(content_); }
-      bool operator==(required_layer_name const&) const noexcept = default;
+      operator T const&() const
+      {
+        if (!content_.has_value()) {
+          throw std::logic_error(
+            std::format("Cannot retrieve layer from product_selector with no layer\n{}\n",
+                        std::stacktrace::current()));
```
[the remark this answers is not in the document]
beojan: That should be left to the FORM developers.

phlex/core/product_selector.hpp
```diff
@@ -42,42 +44,58 @@ namespace phlex {
...
     // The required_layer_name has to be a template for static_assert(false)
     template <std::same_as<experimental::identifier> T>
-    class required_layer_name {
+    class layer_name {
     public:
-      required_layer_name()
-      {
-        static_assert(false, "The layer name has not been set in this product_selector.");
-      }
+      layer_name() : content_(std::nullopt) {}
```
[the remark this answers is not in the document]
beojan: Resolved.

phlex/core/products_consumer.cpp
```diff
@@ -16,11 +24,33 @@ namespace phlex::detail {
...
-                                       product_selectors input_products) :
+                                       product_selectors input_products,
+                                       layers_required layers_required) :
     consumer{std::move(name), std::move(predicates)},
     input_products_{std::move(input_products)},
     layers_{layers_from(input_products_)}
   {
+    using namespace phlex::experimental::literals;
+    if (layers_required == layers_required::always ||
+        (layers_required != layers_required::never && input_products_.size() > 1)) {
```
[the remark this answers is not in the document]
beojan: Let's leave this for another PR, along with optional creator support in Python.

phlex/core/product_selector.cpp
```diff
@@ -50,7 +50,7 @@ namespace phlex {
     if (!match(spec)) {
       return false;
     }
-    if (experimental::identifier(this->layer) != layer) {
+    if (this->layer && experimental::identifier(this->layer) != layer) {
```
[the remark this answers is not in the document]
beojan: Resolved

phlex/core/products_consumer.cpp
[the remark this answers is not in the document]
beojan: Issue created to track this: #818

phlex/core/index_router.cpp
```diff
@@ -197,7 +198,7 @@ namespace phlex::detail {
                                               provider_input_ports_t provider_input_ports)
   {
     for (auto& [input_product, provider_port] : provider_input_ports | std::views::values) {
-      auto [it, _] = index_set_nodes_.emplace(input_product.layer,
+      auto [it, _] = index_set_nodes_.emplace(input_product.layer ? *input_product.layer : "*"_id,
```
knoepfel, line 200: Suggest replacing `"*"_id` with a named value like `layer_wildcard`.

phlex/core/multilayer_join_node.hpp
```diff
@@ -76,8 +76,7 @@ namespace phlex::detail {
       name_{std::move(node_name)},
       layers_{std::move(layer_names)}
     {
-      assert(NInputs == layers_.size());
```
knoepfel, line 79: If this is an invariant that must be true (I think it still is with this PR), then keep the assertion, which expresses the invariant.

phlex/core/multilayer_join_node.hpp
knoepfel: This appears to be an unused using directive.

phlex/core/product_selector.hpp
```diff
@@ -42,42 +50,64 @@ namespace phlex {
         return me.content_.value_or("[ANY]");
       }
       bool operator==(creator_name const&) const noexcept = default;
+      auto operator<=>(creator_name const&) const noexcept = default;
```
knoepfel: Suggest removing `bool operator== ... default;` if `auto operator<=>` is already needed.

phlex/core/product_selector.hpp
```diff
@@ -42,42 +50,64 @@ namespace phlex {
...
-      explicit operator std::string_view() const noexcept { return std::string_view(content_); }
-      bool operator==(required_layer_name const&) const noexcept = default;
+      operator T const&() const
+      {
+        if (!content_.has_value()) {
+          throw std::logic_error(
+            std::format("Cannot retrieve layer from product_selector with no layer\n{}\n",
+                        std::stacktrace::current()));
+        }
+        return content_.operator*();
```
knoepfel: Why not return `*content_`?

phlex/core/CMakeLists.txt
knoepfel, line 1: We should discuss as a group whether it makes sense to add the stacktrace printout to error messages.  That decision is decoupled from making layer specification optional.  Please remove the introduction of the stacktrace library from this PR, and we can discuss it at a dedicated development-team meeting.

phlex/core/product_selector.hpp
```diff
@@ -42,42 +50,64 @@ namespace phlex {
...
       // NOLINTNEXTLINE(google-explicit-constructor) - Implicit conversion is intentional
-      operator T const&() const noexcept { return content_; }
-      explicit operator std::string_view() const noexcept { return std::string_view(content_); }
-      bool operator==(required_layer_name const&) const noexcept = default;
+      operator T const&() const
+      {
+        if (!content_.has_value()) {
+          throw std::logic_error(
+            std::format("Cannot retrieve layer from product_selector with no layer\n{}\n",
+                        std::stacktrace::current()));
```
knoepfel: See above comment about using the stacktrace library.  We'll make that decision at another time.

phlex/core/product_selector.hpp
```diff
@@ -42,42 +50,64 @@ namespace phlex {
...
+        }
+        return content_.operator*();
+      }
+
+      experimental::identifier const& operator*() const noexcept { return content_.operator*(); }
```
knoepfel, line 83: Suggest returning `*content_`.

phlex/core/product_selector.hpp
```diff
@@ -42,42 +50,64 @@ namespace phlex {
...
       }
       bool operator==(creator_name const&) const noexcept = default;
+      auto operator<=>(creator_name const&) const noexcept = default;
 
     private:
       std::optional<experimental::identifier> content_;
     };
 
     // The required_layer_name has to be a template for static_assert(false)
     template <std::same_as<experimental::identifier> T>
```
knoepfel: The `layer_name` no longer needs to a template if there is no longer a `static_assert(false, ....)`.

phlex/core/product_selector.hpp
knoepfel: There no longer appears to be a `static_assert(false, ...)` for this class template, therefore it does not need to be a class template.

phlex/core/product_selector.hpp
```diff
@@ -42,42 +50,64 @@ namespace phlex {
...
+      explicit operator std::string_view() const noexcept
+      {
+        using namespace std::string_view_literals;
+        return content_
+          .transform([](experimental::identifier const& id) { return std::string_view(id); })
+          .value_or(""sv);
+      }
+      operator bool() const noexcept { return content_.has_value(); }
+      bool operator==(layer_name const&) const noexcept = default;
+      auto operator<=>(layer_name const&) const noexcept = default;
```
knoepfel: Suggest removing `bool operator==.... = default;` if the spaceship operator is necessary.

phlex/core/product_selector.hpp
```diff
@@ -42,42 +50,64 @@ namespace phlex {
...
+        return content_.operator*();
+      }
+
+      experimental::identifier const& operator*() const noexcept { return content_.operator*(); }
+      explicit operator std::string_view() const noexcept
+      {
+        using namespace std::string_view_literals;
+        return content_
+          .transform([](experimental::identifier const& id) { return std::string_view(id); })
+          .value_or(""sv);
```
knoepfel: Is there a reason this doesn't return `.value_or("[ANY]"sv)` or `.value_or("*"sv)` so that you do not need to make explicit checks on `layer` in the various places in `product_selector.cpp`?

phlex/core/product_selector.hpp
```diff
@@ -42,42 +50,64 @@ namespace phlex {
...
+      std::optional<experimental::identifier> content_;
     };
   }
 
   struct PHLEX_CORE_EXPORT product_selector {
     detail::creator_name<experimental::identifier> creator;
-    detail::required_layer_name<experimental::identifier> layer;
+    detail::layer_name<experimental::identifier> layer;
     std::optional<experimental::identifier> suffix;
     std::optional<experimental::identifier> stage;
```
knoepfel: The `suffix` and `stage` fields are represented as `std::optional<...>` objects whereas the `creator` and `layer` fields are more strongly typed.  We should discuss whether it makes sense for all of them to be `std::optional<experimental::identifier>` instances, or whether we should have a distinct type for each.  The current implementation straddles these options.

phlex/core/product_selector.cpp
```diff
@@ -77,7 +77,7 @@ namespace phlex {
       suffix.transform(&identifier::operator std::string_view).value_or("[ANY]");
     std::string type_str = this->type.valid() ? fmt::format("<{}>", this->type)
                                               : "[UNSET TYPE]"; // will later be concept
-    auto layer_str = std::string_view(layer);
+    auto layer_str = layer ? std::string_view(layer) : "[ANY]";
```
knoepfel: See comment in `product_selector.hpp` about using `.value_or("[ANY]")` or `.value_or("*")`.

phlex/core/products_consumer.cpp
```diff
@@ -1,13 +1,21 @@
...
+    using namespace phlex::experimental::literals;
     std::vector<phlex::experimental::identifier> result;
     result.reserve(queries.size());
     for (auto const& query : queries) {
-      result.push_back(query.layer);
+      if (query.layer) {
+        result.push_back(query.layer);
+      } else {
+        result.push_back("*"_id);
+      }
```
knoepfel, line 16: A suitable use of `result.push_back(query.layer.value_or(...))` would simplify this code.

phlex/core/products_consumer.cpp
```diff
@@ -1,13 +1,21 @@
...
     }
+    result.shrink_to_fit();
```
knoepfel: This an unnecessary operation: the `result.reserve(...)` above ensures that the allocated vector size is already correct.

phlex/core/products_consumer.hpp
```diff
@@ -17,11 +17,13 @@
 #include <vector>
 
 namespace phlex::detail {
+  enum class layers_required : char { never, multi_input_only, always };
```
knoepfel: Suggest renaming to `require_layers`.  In other words, `require_layers::never` is in the form `[question]::[answer]`, which is easier to read.

phlex/core/products_consumer.hpp
```diff
@@ -17,11 +17,13 @@
...
   class PHLEX_CORE_EXPORT products_consumer : public consumer {
   public:
     products_consumer(phlex::experimental::algorithm_name name,
                       std::vector<std::string> predicates,
-                      product_selectors input_products);
+                      product_selectors input_products,
+                      layers_required layers_required = layers_required::multi_input_only);
```
knoepfel: Suggest removing the default argument.  It is easier for us to remember that (e.g.) `transform_node` needs to be adjusted later on whenever `require_layers::multi_input_only` is supplied.  It also makes it easier to find every place where the `enum` is used.

phlex/core/registrar.hpp
```diff
@@ -112,11 +112,17 @@ namespace phlex::detail {
...
+        auto ptr = creator_(release_predicates(), std::move(output_product_suffixes));
+        auto name = ptr->name().to_string();
+        auto [_, inserted] = nodes_->try_emplace(name, std::move(ptr));
+        if (not inserted) {
+          internal::add_to_error_messages(*errors_, "Node", name);
+        }
+      } catch (...) {
+        // Prevent trying to re-run create_node
+        creator_ = nullptr;
+        throw;
```
knoepfel: Suggest a different approach that avoids the need to re-throw:

```cpp
assert(creator_);
auto create = std::exchange(creator_, node_creator{});
// creator_ now empty

auto ptr = create(...);
...
```

This then permits the removal of `creator_ = nullptr` on line 96.

phlex/core/products_consumer.cpp
knoepfel: This logic is difficult to read.  Suggest creating a function called `validate_layers` with a simpler implementation:

```cpp
void validate_layers(product_selectors const& input_products, require_layers rl)
{
  if (rl == require_layers::never) {
    return;
  }

  if (rl == require_layers::multi_input_only && input_products.size() == 1) {
    return;
  }

  // Now we do the work to see if we're in an error state
}
```

and then the body of the constructor just calls it:

```cpp
products_consumer::products_consumer(...)
{
  validate_layers(input_products_, rl);
}
```

test/product_selecting_test.cpp
```diff
@@ -50,21 +89,21 @@ TEST_CASE("Querying products in different ways", "[graph]")
 
   // Duplicate with transform
   g.transform("duplicate_temperature", [](double const& t) { return t; })
-    .input_family(product_selector{.creator = "input", .layer = "event", .suffix = "temperature"})
+    .input_family(product_selector{.creator = "input", .suffix = "temperature"})
     .output_product_suffixes("temperature");
 
-  SECTION("All fields")
```
knoepfel, line 56: Did we lose a check that does test when all the fields are specified?

knoepfel (dismissed):
@beojan, thank you for the PR.  @marcpaterno and I have finished looking through it and we provide some comments for your consideration.  In addition to our comments, it would be good if the test coverage could be increased.

Marc and I are happy to set up a meeting to discuss things.  Thanks!

phlex/core/multilayer_join_node.hpp
beojan: Done

phlex/core/multilayer_join_node.hpp
beojan: Done

phlex/core/product_selector.hpp
beojan: :heavy_check_mark:

phlex/core/product_selector.hpp
beojan: :heavy_check_mark:

phlex/core/products_consumer.hpp
beojan: :heavy_check_mark:

phlex/core/products_consumer.hpp
beojan: :heavy_check_mark:

phlex/core/registrar.hpp
beojan: Done

phlex/core/product_selector.hpp
beojan: Had to back that change out because it triggers a clang-tidy warning.

phlex/core/product_selector.hpp
beojan: Layer and creator prohibit the empty string.

I don't know if that's really necessary, and worth the extra classes.

phlex/core/products_consumer.cpp
beojan: Done

phlex/core/products_consumer.cpp
beojan: Not with the optional buried in the `layer_name` class. I don't know if it's worth making it public just to simplify this.

phlex/core/index_router.cpp
beojan: Trying to define that somewhere globally accessible produces a clang-tidy warning because the std::string constructor could throw.

test/product_selecting_test.cpp
beojan: The check wasn't really all fields. One that is for all fields has now been added.

beojan:
A few of those comments aren't addressable without triggering new clang-tidy warnings but everything else should be addressed (including coverage).

test/product_selecting_test.cpp
```diff
@@ -73,33 +122,84 @@ TEST_CASE("Querying products in different ways", "[graph]")
...
+    CHECK_THROWS_WITH(
+      g.unfold<copy_temperature_once>("copy_temperature_once",
+                                      &copy_temperature_once::predicate,
+                                      &copy_temperature_once::unfold,
+                                      concurrency::unlimited,
+                                      "temperature_copy")
+        .input_family(product_selector{.creator = "input", .suffix = "temperature"})
+        .output_product_suffixes("temperature"),
+      "Product selectors in layer-mandatory algorithm copy_temperature_once must define their "
+      "layers:\n"
```
knoepfel: I'd suggest changing the error message to something more concise (e.g.):

```
"Must specify layers in the product selectors for node copy_temperature_once:\n"
```

Phlex assigns names to the nodes, not to the algorithms.

knoepfel (commented):
@beojan, this is looking quite good.  Just one more request below.

test/product_selecting_test.cpp
beojan: @codex Can you make this change (both in the code and in the tests)?

beojan: Local codex handled it.

beojan:
@phlexbot format

phlex/core/products_consumer.cpp
knoepfel: Fair enough

phlex/core/product_selector.hpp
knoepfel: Understood

phlex/core/product_selector.hpp
knoepfel: Dismissing comment due to exposing `clang-tidy` unchecked optional access error.

knoepfel (approved):

phlex/core/index_router.cpp
knoepfel: Okay.

phlex/core/product_selector.hpp
knoepfel: Let's leave them for now.

PR #823, opened 2026-08-17
FORM Canonical Index Format

wwuoneway:
**Summary:**
Align the FORM toy test writer/reader with the canonical index format used by Phlex.

The toy writer/reader emitted a hex, `=`/`;`-delimited index string (`[EVENT={:08X};SEG={:08X}]`), which the Phlex source parser (`form_source.cpp::parse_index_string` — base-10, `:`-separated) cannot parse directly. Production write/read already use the canonical form via `phlex::data_cell_index::to_string()`, so the toy files were the only divergent emitter.

This switches the toy writer/reader and the valid index strings in the storage-layer tests to the canonical `[layer:number, ...]` base-10 format, so FORM-written index strings are directly consumable by the source. Intentionally malformed error-path strings are left unchanged — they exercise the reader's rejection logic.

### DIFF
--- a/test/form/form_storage_test.cpp
+++ b/test/form/form_storage_test.cpp
@@ -397,12 +397,11 @@ TEST_CASE("StorageReader getIndex: malformed ids and compatibility fallbacks", "
   CHECK(reader.getIndex(index_token, "plain-text-id", settings) == 0);
 
   // Malformed IDs must be rejected by the storage-layer index parser
-  CHECK_THROWS_AS(reader.getIndex(index_token, "[EVENT,SEG=00000001]", settings),
-                  std::runtime_error);
+  CHECK_THROWS_AS(reader.getIndex(index_token, "[EVENT,SEG=1]", settings), std::runtime_error);
   CHECK_THROWS_AS(
     reader.getIndex(index_token, "[EVENT=99999999999999999999999999999999]", settings),
     std::runtime_error);
-  CHECK_THROWS_AS(reader.getIndex(index_token, "[=00000001]", settings), std::runtime_error);
+  CHECK_THROWS_AS(reader.getIndex(index_token, "[=1]", settings), std::runtime_error);
   CHECK_THROWS_AS(reader.getIndex(index_token, "[EVENT]", settings), std::runtime_error);
   CHECK_THROWS_AS(reader.getIndex(index_token, "[    ]", settings), std::runtime_error);
 }
--- a/test/form/form_storage_test.cpp
+++ b/test/form/form_storage_test.cpp
@@ -64,7 +64,7 @@ TEST_CASE("Storage_Container read wrong type", "[form]")
 TEST_CASE("Storage_Container sharing an Association", "[form]")
 {
   std::vector<float> piData(10, std::numbers::pi_v<float>);
-  std::string indexData = "[EVENT=00000001;SEG=00000001]";
+  std::string indexData = "[event:1, segment:1]";
 
   form::test::write(technology, piData, indexData);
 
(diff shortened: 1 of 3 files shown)


test/form/reader.cpp
```diff
@@ -89,7 +89,8 @@ int main(int argc, char** argv)
     for (int nseg = 0; nseg < NUMBER_SEGMENT; nseg++) {
 
       void const* rawPtr = nullptr;
-      std::string const seg_id_text = std::format("[EVENT={:08X};SEG={:08X}]", nevent, nseg);
+      // Must match the canonical format emitted by writer.cpp (phlex data_cell_index format)
+      std::string const seg_id_text = std::format("[event:{}, segment:{}]", nevent, nseg);
```
pcanal, line 93: Why is it not:

```suggestion
      std::string const seg_id_text = std::format("[event:8, segment:8]", nevent, nseg);
```

aolivier23 (dismissed):
LGTM.  I think I understand what your answer to the outstanding question will be.  Please ping me for another review if it triggers another change.

test/form/reader.cpp
wwuoneway: The suggested `[event:8, segment:8]` has no `{}` placeholders, so `std::format` would ignore `nevent` and `nseg` and emit the literal string `[event:8, segment:8]` for every segment.
If the intent was to zero-pad to width 8 (like the old {:08X}), the syntax would be `{:08}`. I deliberately didn't do that, so that the canonical index format is consistent with `data_cell_index::to_string()`(see `phlex/model/data_cell_index.cpp`). So `[event:{}, segment:{}]` keeps us byte-identical to Phlex. I'll leave it as-is unless you see a reason Phlex should start padding.

test/form/form_storage_test.cpp
```diff
@@ -396,6 +396,7 @@ TEST_CASE("StorageReader getIndex: malformed ids and compatibility fallbacks", "
   CHECK(reader.getIndex(index_token, "[]", settings) == 0);
   CHECK(reader.getIndex(index_token, "plain-text-id", settings) == 0);
 
+  // Malformed IDs must be rejected by the storage-layer index parser
   CHECK_THROWS_AS(reader.getIndex(index_token, "[EVENT,SEG=00000001]", settings),
```
gemmeren: Should this be updated as well? removing leading '0's?

gemmeren (dismissed):
One minor item...
Thanks Wanwei

test/form/form_storage_test.cpp
wwuoneway: These are deliberately-malformed inputs, so their formatting intentionally doesn't follow the canonical shape — they exist to exercise the parser's rejection paths.

test/form/reader.cpp
pcanal: Thanks for the clarification (I did indeed misread the code).

pcanal (dismissed):
Thanks.

test/form/form_storage_test.cpp
gemmeren: I understand. But they were deliberately malformed by dropping the even-number. Now having leading '0''s on the sequence number is just odd. 

PS: If this test is meant to detect throwing when event number is missing, it could get hidden by it throwing because of leading '0's in segment. Likely it can digest the '0's, but why have them here, they are a leftover of the previous format and do no longer apply, but messy up the test of detecting missing id pieces.

wwuoneway: Good point — Done as suggested.

gemmeren (approved):

PR #827, opened 2026-08-18
fix(cmake): propagate public dependencies to installed package config

### LINKED ISSUE #656: Installed `phlexConfig.cmake` is missing `find_dependency()` calls for public dependencies
I tried building a minimal cmake package against phlex and found I had to include dependencies in my package that are not directly used in order to satisfy what phlex requires.  I believe the fix is for Phlex's cmake should add some `find_dependency()` calls.

I hope it is okay to paste details generated from claude.  If this kind of content is objectionable for the Phlex project, I'll hand-craft any future issue texts.  

Or, if you prefer a PR, I can make one, though I am not a cmake expert.

Cheers,
-Brett.

Now, Claude writes:

## Summary

The installed package config (`phlexConfig.cmake`) does not call `find_dependency()`
for phlex's public dependencies. As a result, the imported targets it defines
reference other packages' targets (`Boost::boost`, `spdlog::spdlog`, `fmt::fmt`,
`TBB::tbb`, `Boost::json`) that are not guaranteed to exist in the consumer's
project, so `find_package(phlex)` can succeed yet leave the targets unusable.

## Details

`phlexTargets.cmake` defines imported targets whose `INTERFACE_LINK_LIBRARIES`
name targets from other packages, e.g.:

```cmake
# phlex::model
INTERFACE_LINK_LIBRARIES "Boost::boost;spdlog::spdlog;fmt::fmt"
# phlex::core
INTERFACE_LINK_LIBRARIES "TBB::tbb;phlex::metaprogramming_int;phlex::model;phlex::utilities"
```

But `phlexConfig.cmake` includes `CMakeFindDependencyMacro` and then makes **no**
`find_dependency()` calls before including the targets file:

```cmake
include(CMakeFindDependencyMacro)   # included, but never used
...
include("${CMAKE_CURRENT_LIST_DIR}/phlexTargets.cmake")
```

The contract of a config package is that after `find_package(phlex)` succeeds,
every imported target it defines is usable. Today that only holds if the
consumer happens to have already called `find_package(Boost)`,
`find_package(spdlog)`, `find_package(TBB)`, etc. themselves.

## Symptom

A minimal consumer that links, say, only `phlex::model`:

```cmake
find_package(phlex REQUIRED)
target_link_libraries(mylib PUBLIC phlex::model)
```

fails at configure time with:

```
CMake Error at .../phlexTargets.cmake:85 (set_target_properties):
  The link interface of target "phlex::model" contains:
    Boost::boost
  but the target was not found.
```

The current workaround is for every consumer to replicate phlex's internal
dependency list (`find_package(Boost)`, `find_package(spdlog)`, ...) before
`find_package(phlex)`. That is brittle: when phlex changes what its targets
link, every consumer silently breaks until updated.

## Fix

Re-find phlex's public dependencies in `phlexConfig.cmake.in`, *before* including
the targets file, using `find_dependency()` (which forwards `QUIET`/`REQUIRED`
and fails with a clear message):

```cmake
include(CMakeFindDependencyMacro)
find_dependency(Boost)
find_dependency(spdlog)   # transitively finds fmt
find_dependency(TBB)
# ... any other public deps (Boost components such as Boost::json) ...

include("${CMAKE_CURRENT_LIST_DIR}/phlexTargets.cmake")
```

Match the version constraints / components to what phlex actually requires
(e.g. specific Boost components). Once this is in place, consumers can drop
their workaround `find_package()` calls entirely.

### Alternative

For any of these that are *implementation details* not exposed in phlex's public
headers, the cleaner fix is to make them `PRIVATE` link dependencies so they
never appear in `INTERFACE_LINK_LIBRARIES` and consumers need not find them at
all. Dependencies whose types appear in public phlex headers are genuinely
public and should be handled via `find_dependency()` as above.

knoepfel:
Resolves #656.

- **Build system**
  - Split dependencies into public API and internal implementation groups.
  - Export Boost JSON, TBB, fmt, and spdlog so the installed `phlexConfig.cmake` can re-find them with `find_dependency()`.
  - Keep Boost program_options and jsonnet internal because they are not part of the public dependency surface.
  - Ensure consumers can use installed Phlex targets without declaring Phlex’s public dependencies separately.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -158,11 +158,15 @@ set(CMAKE_INSTALL_RPATH_USE_LINK_PATH ON)
 
 list(PREPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/Modules)
 
-find_package(Boost REQUIRED COMPONENTS json program_options)
-find_package(TBB REQUIRED)
-find_package(fmt REQUIRED)
+# Dependencies required by public API
+find_package(Boost REQUIRED COMPONENTS json EXPORT)
+find_package(TBB REQUIRED EXPORT)
+find_package(fmt REQUIRED EXPORT)
+find_package(spdlog REQUIRED EXPORT)
+
+# Dependencies required by internal implementation
+find_package(Boost REQUIRED COMPONENTS program_options)
 find_package(jsonnet REQUIRED)
-find_package(spdlog REQUIRED)
 
 # Apply ThreadSanitizer flags if enabled
 if(ENABLE_TSAN)


knoepfel:
@coderabbit full review

knoepfel:
@CodeRabbit full review

greenc-FNAL (approved):

PR #829, opened 2026-08-19
fix(core): preserve wide counts in indexed end tokens

### LINKED ISSUE #810: Prevent narrowing committed counts in indexed end tokens
## Summary

`index_router` converts a committed count from `std::size_t` to `int` when it sends an `indexed_end_token`. Counts greater than `INT_MAX` can narrow to an implementation-defined value. A negative value can then cause incorrect subtraction in flush-token handling.

## Required changes

- Change `indexed_end_token::count` and the corresponding flush-token handling to use `std::ptrdiff_t`.
- Update `repeater_node::handle_flush_token` and related code to preserve the widened count type.
- Remove the `static_cast<int>(count)` conversion in the `index_router::route` flush callback.
- Preserve the existing count propagation behavior.

## Rationale

A fold partition can accumulate more than `INT_MAX` cells. The current narrowing conversion can silently corrupt the committed count before `repeater_node::handle_flush_token` processes it.

## Affected areas

- `phlex/core/index_router.cpp`
- The `indexed_end_token` declaration and its producers/consumers
- `repeater_node::handle_flush_token`

## Acceptance criteria

- No committed count is narrowed from `std::size_t` to `int` during end-token creation.
- `indexed_end_token` and all flush-token consumers use `std::ptrdiff_t` consistently.
- Existing flush-token count propagation semantics remain unchanged.
- Add or update tests for counts above `INT_MAX`, where practical.

## Backlinks

- Pull request: https://github.com/Framework-R-D/phlex/pull/540
- Review comment: https://github.com/Framework-R-D/phlex/pull/540#discussion_r3724468141
- Requested by: @knoepfel

knoepfel:
Resolves #810 

- **Code**
  - Widen indexed end-token and flush-token counts from `int` to `std::ptrdiff_t`.
  - Preserve wide counts through `flush_gate`, `repeater_node`, and `index_router`.
  - Remove the narrowing `static_cast<int>(count)` conversion.
  - Update accumulator and repeater counters to use `signed_size_t`.
  - Test behavior 

- **Behavior**
  - Preserve existing count propagation.
  - Support committed counts greater than `INT_MAX` without narrowing.
  - Throw when the expected count total exceeds the maximum value of `signed_size_t`.

### DIFF
--- a/phlex/core/detail/accumulator_node.hpp
+++ b/phlex/core/detail/accumulator_node.hpp
@@ -4,6 +4,7 @@
 #include "phlex/core/fold/send.hpp"
 #include "phlex/core/message.hpp"
 #include "phlex/phlex_core_export.hpp"
+#include "phlex/utilities/signed_size.hpp"
 
 #include "oneapi/tbb/concurrent_hash_map.h"
 #include "oneapi/tbb/concurrent_queue.h"
@@ -122,7 +123,9 @@ namespace phlex::detail::internal {
     struct cached_accumulator {
       std::shared_ptr<accumulator_msg_t> accumulator_msg;
       tbb::concurrent_queue<std::size_t> msg_ids;
-      std::atomic<std::ptrdiff_t> counter;
+      // Signed balance of pending invocations. It may be negative when a flush arrives before
+      // all concurrent invocations are processed; zero means that the partition is complete.
+      std::atomic<signed_size_t> pending_invocations;
       std::atomic_flag flush_received;
       std::size_t original_message_id{};
     };
@@ -271,7 +274,7 @@ namespace phlex::detail::internal {
     entry->original_message_id = msg.msg_id;
     emit_pending_ids(entry);
     // Handle the flush-before-partition case: if the flush token already arrived (and
-    // left counter == 0 because no fold inputs flowed under this partition), emit the
+    // left pending_invocations == 0 because no fold inputs flowed under this partition), emit the
     // initial accumulator value now that we finally have the index and original message
     // ID needed to construct the output.
     cleanup_cache_entry(a);
@@ -284,7 +287,7 @@ namespace phlex::detail::internal {
     accessor a;
     cached_results_.insert(a, index->hash());
     auto* entry = &a->second;
-    entry->counter -= count;
+    entry->pending_invocations -= count;
     std::ignore = entry->flush_received.test_and_set();
     cleanup_cache_entry(a);
   }
@@ -313,10 +316,11 @@ namespace phlex::detail::internal {
     auto* entry = &a->second;
     // The `accumulator_msg` check guards the flush-before-partition case: a zero-count
     // flush can establish the cache entry with `flush_received == true` and
-    // `counter == 0` before the partition message has supplied the index and initial
+    // `pending_invocations == 0` before the partition message has supplied the index and initial
     // value.  In that case we defer emission until `handle_partition_message` re-enters
     // cleanup with a populated `accumulator_msg`.
-    if (entry->flush_received.test() and entry->counter == 0 and entry->accumulator_msg) {
+    if (entry->flush_received.test() and entry->pending_invocations == 0 and
+        entry->accumulator_msg) {
       output_port<0>(repeater_).try_put(entry->accumulator_msg->release_as_message(
         node_name_, output_, entry->original_message_id));
       ++emitted_result_count_;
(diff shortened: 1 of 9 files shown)


phlex/core/index_router.cpp
```diff
@@ -282,7 +282,7 @@ namespace phlex::detail {
       [end_token_entries = std::move(end_token_entries)](flush_gate const& fc) {
         for (auto const& entry : *end_token_entries) {
           auto const count = fc.committed_count_for_layer(entry.counting_layer_hash);
-          entry.flush_port->try_put({.index = fc.index(), .count = static_cast<int>(count)});
+          entry.flush_port->try_put({.index = fc.index(), .count = count});
```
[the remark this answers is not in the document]
knoepfel: @coderabbit, I'm not sure that a regression test is necessary here.  The `static_cast<int>` was being performed due to type mismatches.  If we still have a type mismatch somewhere, then I agree that the current solution should be tested to make sure the mismatch is appropriately handled.  However, if there are no longer any type mismatches, then adding a regression test seems unnecessary.  Please advise.

phlex/core/detail/accumulator_node.hpp
```diff
@@ -121,7 +122,7 @@ namespace phlex::detail::internal {
     struct cached_accumulator {
       std::shared_ptr<accumulator_msg_t> accumulator_msg;
       tbb::concurrent_queue<std::size_t> msg_ids;
-      std::atomic<int> counter;
+      std::atomic<std::ptrdiff_t> counter;
```
pcanal: Why was the counter a signed number? i.e. why not:
```suggestion
      std::atomic<std::size_t> counter;
```

knoepfel: The counter is signed because a count of 0 indicates the accumulation is complete:

- `counter += 1` for each expected invocation of a fold operator
- `counter -= n` whenever the total number `n` of expected invocations is known

Because the total number `n` may arrive before all fold-operator invocations are complete (thanks to concurrency), the counter value may go negative.

An alternative formulation is to keep track of separate numbers: the number of completed invocations and the number of expected invocations.  An older implementation used this method, thus avoiding potential narrowing, but introducing significant implementation complexity.

phlex/core/index_router.cpp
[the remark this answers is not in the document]
knoepfel: @coderabbit, see latest commit.

phlex/core/detail/accumulator_node.hpp
pcanal: Should we consider adding (a version of) this information as a code comment for the data member `counter` (which unlike its peer is really not self describing) ?

knoepfel: Yes, we should describe this.  Stay tuned.

pcanal:
```suggestion
      std::atomic<std::make_signed<size_t>> counter;
```
In practice (for our platforms) this does not make an actual difference but the C++ does not guarantee that `sizeof(ptrdiff_t) == sizeof(size_t)`.

pcanal (dismissed):
LGTM.

phlex/core/detail/accumulator_node.hpp
knoepfel: Actually, I like the idea of using `std::make_signed<std::size_t>`.  Will adopt.

knoepfel: See commit https://github.com/Framework-R-D/phlex/pull/829/commits/57315e8f8eab84dfff55ea4c4b3cbe22a263999e.

knoepfel: See commit https://github.com/Framework-R-D/phlex/pull/829/commits/57315e8f8eab84dfff55ea4c4b3cbe22a263999e.

phlex/core/message.hpp
```diff
@@ -31,7 +32,7 @@ namespace phlex::detail {
 
   struct indexed_end_token {
     data_cell_index_ptr index;
-    int count;
+    signed_size_t count;
```
pcanal, line 39: Can we add a code comment with the semantic of this count?

knoepfel: @pcanal, see commit https://github.com/Framework-R-D/phlex/pull/829/commits/de5c1bcae079256546719258c3986209b2301632.

pcanal (approved):
Thanks.

PR #830, opened 2026-08-19
Form:  Return a Token from registerWrite

wwuoneway:
**Why:** When a product is written we currently get nothing back, so the row it landed in is only recoverable by re-scanning the container on read. This surfaces that information at write time as a Token (placement + row) — the piece needed to locate the product again, and the basis for the upcoming navigation/index work (FORM will collect these Tokens and hand them to commitOutput()).

**What:**

- `registerWrite()` now returns a **Token** built directly from the product's placement and the 0-based row it was written to. If the backend doesn't address rows, the Token has no id set.

- The row / Token::id type is `std::uint64_t` — wide enough for both backends' entry counts (TBranch::GetEntries(), RNTuple::GetNEntries()) and never narrower than RNTuple. Threaded through the write chain: `fill` / `fillContainer` / `registerWrite`. 

- Token gains an explicit not-set state (`hasId()`), replacing the old -1 sentinel. The write chain uses `kInvalidRowId` for "no addressable row"; `registerWrite` maps it to a not-set Token.

- ROOT backends: TBranch returns `GetEntries()` - 1 (row persists on Fill()); RField returns `GetNEntries()` (entry persists on commit()). Both yield the same 0-based row the reader uses.

**Note of follow-up:**

- The read path stays `int` for now; the single write→read narrowing is made explicit with a documented static_cast<int>. Widening the read path to 64-bit is a follow-up.
-  FORM will collects Tokens → `commitOutput()` registers them into an index container, which is a follow-up.

- **Code**
  - Updated `registerWrite()` to return a `Token` with product placement and a 0-based row ID.
  - Changed write-side row IDs and `Token::id()` to `std::uint64_t`.
  - Added `Token::hasId()` and replaced the `-1` sentinel with an explicit unset state.
  - Added `kInvalidRowId` for backends that do not address rows.
  - Propagated row IDs through `fill`, `fillContainer`, and `registerWrite`.
  - Aligned ROOT row handling with reader-visible rows:
    - `TBranch` uses `GetEntries() - 1`.
    - `RField` uses `GetNEntries()`.
  - Added an explicit narrowing cast when the read path passes the token row ID to the existing `int` API.

- **Tests**
  - Updated `Token` tests for `hasId()` and unsigned IDs.
  - Added tests for returned tokens, sequential row IDs, direct token-based reads, and non-addressable backends.

- **Deferred**
  - FORM token collection and widening of the read path remain follow-up work.

### DIFF
--- a/form/core/token.hpp
+++ b/form/core/token.hpp
@@ -5,6 +5,7 @@
 
 #include "core/technology.hpp"
 
+#include <cstdint>
 #include <string>
 
 /* @class Token
@@ -13,11 +14,17 @@
 namespace form::detail::experimental {
   class Token {
   public:
-    /// Default constructor; delegates to the named constructor so the -1 sentinel for id is defined once
+    /// Default constructor; a token with no id set (delegates to the placement-only constructor)
     Token() : Token("", "", {}) {}
 
-    /// Named constructor; id defaults to -1 as a "not set" sentinel
-    Token(std::string fileName, std::string containerName, technology::Id technology, int id = -1);
+    /// Placement-only constructor; leaves the id unset (hasId() == false)
+    Token(std::string fileName, std::string containerName, technology::Id technology);
+
+    /// Fully-specified constructor; sets the 0-based row/entry id (hasId() == true)
+    Token(std::string fileName,
+          std::string containerName,
+          technology::Id technology,
+          std::uint64_t id);
 
     /// Access file name
     std::string const& fileName() const;
@@ -26,8 +33,10 @@ namespace form::detail::experimental {
     /// Access technology type
     technology::Id technology() const;
 
-    /// Access identifier/entry number
-    int id() const;
+    /// Access identifier/entry number (0-based row). Only meaningful when hasId() is true.
+    std::uint64_t id() const;
+    /// Whether an id has been set on this token
+    bool hasId() const;
 
   private:
     /// Technology identifier
@@ -36,8 +45,10 @@ namespace form::detail::experimental {
     std::string m_fileName;
     /// Container name
     std::string m_containerName;
-    /// Identifier/entry number
-    int m_id;
+    /// Identifier/entry number (0-based row)
+    std::uint64_t m_id;
+    /// Whether m_id holds a valid, set value
+    bool m_hasId;
   };
 } // namespace form::detail::experimental
 #endif // FORM_CORE_TOKEN_HPP
--- a/form/persistence/persistence_writer.cpp
+++ b/form/persistence/persistence_writer.cpp
@@ -48,13 +48,13 @@ void PersistenceWriter::createContainers(
   m_store_writer->createContainers(containers, m_tech_settings);
 }
 
-void PersistenceWriter::registerWrite(std::string const& creator,
-                                      std::string const& label,
-                                      void const* data,
-                                      std::type_info const& type)
+int PersistenceWriter::registerWrite(std::string const& creator,
+                                     std::string const& label,
+                                     void const* data,
+                                     std::type_info const& type)
 {
   std::unique_ptr<Placement> plcmnt = getPlacement(creator, label);
-  m_store_writer->fillContainer(*plcmnt, data, type);
+  return m_store_writer->fillContainer(*plcmnt, data, type);
 }
 
 void PersistenceWriter::commitOutput(std::string const& creator, std::string const& id)
--- a/form/persistence/persistence_writer.cpp
+++ b/form/persistence/persistence_writer.cpp
@@ -55,11 +55,12 @@ Token PersistenceWriter::registerWrite(std::string const& creator,
 {
   std::unique_ptr<Placement> plcmnt = getPlacement(creator, label);
   std::uint64_t const row = m_store_writer->fillContainer(*plcmnt, data, type);
-  // kInvalidRowId is not an error: it means the backend does not address rows
-  // (e.g. the generic no-op container). Real write failures throw in the backend.
-  // Represent "no addressable row" as a Token with hasId() == false.
+  // A returned Token must locate a readable product: its row is the read-side navigation key.
+  // kInvalidRowId means backend does not address rows,so a product routed there could not be located on read, so throw here for such an unusable Token
   if (row == kInvalidRowId) {
-    return Token{plcmnt->fileName(), plcmnt->containerName(), plcmnt->technology()};
+    throw std::runtime_error("PersistenceWriter::registerWrite backend for product '" + label +
+                             "' from creator '" + creator + "' does not address rows; " +
+                             "cannot produce a Token locating the written product");
   }
   return Token{plcmnt->fileName(), plcmnt->containerName(), plcmnt->technology(), row};
 }
--- a/form/persistence/persistence_writer.cpp
+++ b/form/persistence/persistence_writer.cpp
@@ -48,13 +48,17 @@ void PersistenceWriter::createContainers(
   m_store_writer->createContainers(containers, m_tech_settings);
 }
 
-int PersistenceWriter::registerWrite(std::string const& creator,
-                                     std::string const& label,
-                                     void const* data,
-                                     std::type_info const& type)
+Token PersistenceWriter::registerWrite(std::string const& creator,
+                                       std::string const& label,
+                                       void const* data,
+                                       std::type_info const& type)
 {
   std::unique_ptr<Placement> plcmnt = getPlacement(creator, label);
-  return m_store_writer->fillContainer(*plcmnt, data, type);
+  std::uint64_t const row = m_store_writer->fillContainer(*plcmnt, data, type);
+  if (row == kInvalidRowId) {
+    return Token{plcmnt->fileName(), plcmnt->containerName(), plcmnt->technology()};
+  }
+  return Token{plcmnt->fileName(), plcmnt->containerName(), plcmnt->technology(), row};
 }
 
 void PersistenceWriter::commitOutput(std::string const& creator, std::string const& id)
--- a/form/persistence/persistence_writer.cpp
+++ b/form/persistence/persistence_writer.cpp
@@ -55,6 +55,9 @@ Token PersistenceWriter::registerWrite(std::string const& creator,
 {
   std::unique_ptr<Placement> plcmnt = getPlacement(creator, label);
   std::uint64_t const row = m_store_writer->fillContainer(*plcmnt, data, type);
+  // kInvalidRowId is not an error: it means the backend does not address rows
+  // (e.g. the generic no-op container). Real write failures throw in the backend.
+  // Represent "no addressable row" as a Token with hasId() == false.
   if (row == kInvalidRowId) {
     return Token{plcmnt->fileName(), plcmnt->containerName(), plcmnt->technology()};
   }
--- a/form/root_storage/root_rfield_write_container.cpp
+++ b/form/root_storage/root_rfield_write_container.cpp
@@ -61,7 +61,7 @@ namespace form::detail::experimental {
     m_rntuple_parent = parentDerived;
   }
 
-  int ROOT_RField_Write_ContainerImp::fill(void const* data)
+  std::uint64_t ROOT_RField_Write_ContainerImp::fill(void const* data)
   {
     if (!m_rntuple_parent) {
       throw std::runtime_error(
@@ -83,7 +83,7 @@ namespace form::detail::experimental {
     // Unlike a TBranch, an RNTuple entry is only written on commit();
     // every field bound before that commit shares one entry.
     // Return the 0-based index that pending entry will occupy (the current entry count).
-    return static_cast<int>(m_rntuple_parent->m_writer->GetNEntries());
+    return static_cast<std::uint64_t>(m_rntuple_parent->m_writer->GetNEntries());
   }
 
   void ROOT_RField_Write_ContainerImp::commit()
--- a/form/root_storage/root_rfield_write_container.cpp
+++ b/form/root_storage/root_rfield_write_container.cpp
@@ -61,7 +61,7 @@ namespace form::detail::experimental {
     m_rntuple_parent = parentDerived;
   }
 
-  void ROOT_RField_Write_ContainerImp::fill(void const* data)
+  int ROOT_RField_Write_ContainerImp::fill(void const* data)
   {
     if (!m_rntuple_parent) {
       throw std::runtime_error(
@@ -79,6 +79,11 @@ namespace form::detail::experimental {
       m_rntuple_parent->m_entry = m_rntuple_parent->m_writer->CreateRawPtrWriteEntry();
     }
     m_rntuple_parent->m_entry->BindRawPtr(col_name(), data);
+
+    // Unlike a TBranch, an RNTuple entry is only written on commit();
+    // every field bound before that commit shares one entry.
+    // Return the 0-based index that pending entry will occupy (the current entry count).
+    return static_cast<int>(m_rntuple_parent->m_writer->GetNEntries());
   }
 
   void ROOT_RField_Write_ContainerImp::commit()
--- a/form/root_storage/root_tbranch_write_container.cpp
+++ b/form/root_storage/root_tbranch_write_container.cpp
@@ -94,7 +94,7 @@ void ROOT_TBranch_Write_ContainerImp::setupWrite(std::type_info const& type)
   }
 }
 
-int ROOT_TBranch_Write_ContainerImp::fill(void const* data)
+std::uint64_t ROOT_TBranch_Write_ContainerImp::fill(void const* data)
 {
   // NOTE: incoming parameter `data` is `const` due to the constraints on how we
   // expect users to interact with the data; however, ROOT's SetBranchAddress
@@ -115,8 +115,9 @@ int ROOT_TBranch_Write_ContainerImp::fill(void const* data)
   m_branch->Fill();
   m_branch->ResetAddress();
 
-  // 0-based entries: GetEntries() is the total count after this Fill(); row = count - 1
-  return static_cast<int>(m_branch->GetEntries()) - 1;
+  // 0-based entries: GetEntries() is the total count after this Fill(); row = count - 1.
+  // GetEntries() >= 1 here (we just filled), so the row is non-negative.
+  return static_cast<std::uint64_t>(m_branch->GetEntries() - 1);
 }
 
 void ROOT_TBranch_Write_ContainerImp::commit()
--- a/form/root_storage/root_tbranch_write_container.cpp
+++ b/form/root_storage/root_tbranch_write_container.cpp
@@ -112,11 +112,17 @@ std::uint64_t ROOT_TBranch_Write_ContainerImp::fill(void const* data)
   } else {
     m_branch->SetAddress(reinterpret_cast<void*>(&data));
   }
-  m_branch->Fill();
+  // TBranch::Fill() returns the number of bytes committed, or a negative value on a write error.
+  // ROOT increments entry count before a basket write can fail, so check return value first
+  Int_t const nbytes = m_branch->Fill();
   m_branch->ResetAddress();
+  if (nbytes < 0) {
+    throw std::runtime_error("ROOT_TBranch_Write_ContainerImp::fill TBranch::Fill() failed for " +
+                             col_name());
+  }
 
   // 0-based entries: GetEntries() is the total count after this Fill(); row = count - 1.
-  // GetEntries() >= 1 here (we just filled), so the row is non-negative.
+  // GetEntries() >= 1 here (Fill() succeeded), so the row is non-negative.
   return static_cast<std::uint64_t>(m_branch->GetEntries() - 1);
 }
 
--- a/form/root_storage/root_tbranch_write_container.cpp
+++ b/form/root_storage/root_tbranch_write_container.cpp
@@ -94,7 +94,7 @@ void ROOT_TBranch_Write_ContainerImp::setupWrite(std::type_info const& type)
   }
 }
 
-void ROOT_TBranch_Write_ContainerImp::fill(void const* data)
+int ROOT_TBranch_Write_ContainerImp::fill(void const* data)
 {
   // NOTE: incoming parameter `data` is `const` due to the constraints on how we
   // expect users to interact with the data; however, ROOT's SetBranchAddress
@@ -114,6 +114,9 @@ void ROOT_TBranch_Write_ContainerImp::fill(void const* data)
   }
   m_branch->Fill();
   m_branch->ResetAddress();
+
+  // 0-based entries: GetEntries() is the total count after this Fill(); row = count - 1
+  return static_cast<int>(m_branch->GetEntries()) - 1;
 }
 
 void ROOT_TBranch_Write_ContainerImp::commit()
--- a/form/storage/istorage.hpp
+++ b/form/storage/istorage.hpp
@@ -41,9 +41,10 @@ namespace form::detail::experimental {
     virtual void createContainers(
       std::map<std::unique_ptr<Placement>, std::type_info const*> const& containers,
       form::experimental::config::tech_setting_config const& settings) = 0;
-    virtual void fillContainer(Placement const& plcmnt,
-                               void const* data,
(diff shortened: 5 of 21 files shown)


form/storage/storage_reader.cpp
```diff
@@ -388,5 +388,6 @@ void StorageReader::readContainer(Token const& token,
       cont->second->setAttribute(key, value);
     }
   }
-  cont->second->read(token.id(), data, type);
+  // TODO: Token::id() is a 64-bit row; the read container interface still takes an int entry. Narrow explicitly here (exact for all realistic row counts). Widening the read path to 64-bit is a follow-up PR.
+  cont->second->read(static_cast<int>(token.id()), data, type);
```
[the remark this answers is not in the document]
wwuoneway: This is noted in the description and as TODO, will be a follow-up on the read side.

form/core/token.hpp
```diff
@@ -26,8 +33,10 @@ namespace form::detail::experimental {
     /// Access technology type
     technology::Id technology() const;
 
-    /// Access identifier/entry number
-    int id() const;
+    /// Access identifier/entry number (0-based row). Only meaningful when hasId() is true.
+    std::uint64_t id() const;
+    /// Whether an id has been set on this token
+    bool hasId() const;
```
aolivier23, line 39: Returning `std::optional<std::uint64_t>` might get us out of needing the sentinel value.  The up-side is that we eliminate a very unlikely edge case where someone write a file with 2^64 entries in it (yikes!) and our code "looks" more modern to outside reviewers.  The down-side is that a lot of our peers probably don't know how `std::optional<>` works and will find it harder to read.  I'm certain Kyle, et al. do.  But a new grad student debugging some workflow won't.

form/root_storage/root_rfield_write_container.cpp
```diff
@@ -61,7 +61,7 @@ namespace form::detail::experimental {
     m_rntuple_parent = parentDerived;
   }
 
-  void ROOT_RField_Write_ContainerImp::fill(void const* data)
+  std::uint64_t ROOT_RField_Write_ContainerImp::fill(void const* data)
```
aolivier23, line 64: Looks great to me!  My only very slight concern is that I think we could get a "corrupt" Token if an exception causes the job to shut down between `fill()` and `commit()`.  Error handling like this is probably a topic for a future PR/issue anyway.

form/storage/storage_write_container.cpp
```diff
@@ -18,7 +18,7 @@ void Storage_Write_Container::setFile(std::shared_ptr<IStorage_File> file) { m_f
 
 void Storage_Write_Container::setupWrite(std::type_info const& /* type*/) {}
 
-void Storage_Write_Container::fill(void const* /* data*/) {}
+std::uint64_t Storage_Write_Container::fill(void const* /* data*/) { return kInvalidRowId; }
```
aolivier23, line 21: `std::optional<>` might fit in well here too.  Very slightly better readability for young experts at the cost of slightly worse readability for our broadest possible audience.

test/form/form_storage_test.cpp
```diff
@@ -331,6 +331,91 @@ TEST_CASE("Persistence round-trip: structured index normalization and listing",
...
+  reader.readContainer(token_first, &raw, typeid(std::vector<int>), settings);
+  auto const* got_first = static_cast<std::vector<int> const*>(raw);
+  REQUIRE(got_first != nullptr);
+  CHECK(*got_first == first);
+
+  raw = nullptr;
+  reader.readContainer(token_second, &raw, typeid(std::vector<int>), settings);
+  auto const* got_second = static_cast<std::vector<int> const*>(raw);
+  REQUIRE(got_second != nullptr);
+  CHECK(*got_second == second);
```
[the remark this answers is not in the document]
aolivier23: Agreed this is worth doing please.  If I'm reading coderabbit's comment correctly, we'd eliminate a little noise from memory debugging tools like `valgrind`.

aolivier23 (dismissed):
Looks good enough to me.  I could be very happy with the code as-is.  Two additional things we could consider:
- `std::optional<>` instead of sentinel values: worth using the newer standard, or better to use the well-known technique?
- We should fix that coderabbit comment about the test if we can.  I like to memcheck our tests when something goes wrong.

form/core/token.hpp
```diff
@@ -36,8 +45,10 @@ namespace form::detail::experimental {
     std::string m_fileName;
     /// Container name
     std::string m_containerName;
-    /// Identifier/entry number
-    int m_id;
+    /// Identifier/entry number (0-based row)
+    std::uint64_t m_id;
+    /// Whether m_id holds a valid, set value
+    bool m_hasId;
   };
```
pcanal, line 52: Consider
```suggestion
    /// Identifier/entry number (0-based row)
    std::optional<std::uint64_t> m_id;
```

form/root_storage/root_rfield_write_container.cpp
```diff
@@ -79,6 +79,11 @@ namespace form::detail::experimental {
       m_rntuple_parent->m_entry = m_rntuple_parent->m_writer->CreateRawPtrWriteEntry();
     }
     m_rntuple_parent->m_entry->BindRawPtr(col_name(), data);
+
+    // Unlike a TBranch, an RNTuple entry is only written on commit();
+    // every field bound before that commit shares one entry.
+    // Return the 0-based index that pending entry will occupy (the current entry count).
+    return static_cast<std::uint64_t>(m_rntuple_parent->m_writer->GetNEntries());
```
pcanal, line 86: I am guessing that this information is then accurate if and only if the RNTuple never uses the multi-thread writing (to the same RNTuple) ....

form/core/token.hpp
```diff
@@ -13,11 +14,17 @@
...
-    /// Named constructor; id defaults to -1 as a "not set" sentinel
-    Token(std::string fileName, std::string containerName, technology::Id technology, int id = -1);
+    /// Placement-only constructor; leaves the id unset (hasId() == false)
+    Token(std::string fileName, std::string containerName, technology::Id technology);
+
+    /// Fully-specified constructor; sets the 0-based row/entry id (hasId() == true)
+    Token(std::string fileName,
+          std::string containerName,
+          technology::Id technology,
+          std::uint64_t id);
```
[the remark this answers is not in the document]
wwuoneway: Thanks — you're right about the phlex guideline. The catch is that `form/` currently follows its own naming convention (e.g. `hasId`, no trailing underscore on private members) throughout, so renaming just this one constructor + `hasId()`/`m_hasId` here would make this file inconsistent with the rest of the subtree rather than more consistent with phlex.

Rather than do a partial rename in this PR, we'd prefer to align the whole form/ and test/form/ tree in one coordinated pass. That naming question is still under discussion with the team; once it's settled we'll do a single sweeping rename so the convention is applied uniformly.

So I'm going to leave the identifiers as-it in this PR to keep it consistent with its neighbors. Happy to track the tree-wide rename as a separate issue — feel free to open one (or I can), and we'll reference it here.

form/storage/istorage.hpp
```diff
@@ -7,13 +7,19 @@
...
 #include <map>
 #include <memory>
 #include <string>
 #include <vector>
 
 namespace form::detail::experimental {
 
+  // Sentinel returned by the write chain when no addressable row was written
+  // (e.g. the generic no-op container). A real row is always < this value.
+  inline constexpr std::uint64_t kInvalidRowId = std::numeric_limits<std::uint64_t>::max();
```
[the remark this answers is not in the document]
wwuoneway: Similar to above (just copy the reply below):
Thanks — you're right about the phlex guideline. The catch is that form/ currently follows its own naming convention (e.g. hasId, no trailing underscore on private members) throughout, so renaming just this one constructor + hasId()/m_hasId here would make this file inconsistent with the rest of the subtree rather than more consistent with phlex.

Rather than do a partial rename in this PR, we'd prefer to align the whole form/ and test/form/ tree in one coordinated pass. That naming question is still under discussion with the team; once it's settled we'll do a single sweeping rename so the convention is applied uniformly.

So I'm going to leave the identifiers as-it in this PR to keep it consistent with its neighbors. Happy to track the tree-wide rename as a separate issue — feel free to open one, and we'll reference it here.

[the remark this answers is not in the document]
wwuoneway: Yes, please create a GitHub issue for the coordinated `form/` and `test/form/` rename.

form/core/token.hpp
pcanal: `std::optional` does not have to leak out to the interface:
```
bool hasId() const { return m_Id.has_value(); };
```

Related question is whether the behavior of `id()` should be:
```
    /// Access identifier/entry number (0-based row). Only meaningful when hasId() is true and throws when called with `hadId() == false`
```

wwuoneway: Good point, and I agree `std::optional` is attractive for the type-safety (can't read the id without checking) more than the 2^64 case. One clarification: `Token` itself doesn't use a sentinel — it's `uint64_t m_id` + `bool m_hasId`, so there's no id-value collision risk in the token; the `max()` sentinel only lives in the write chain (`fill()`/`fillContainer()` → `kInvalidRowId`). And that one is unreachable in practice: `TBranch::GetEntries()` is a signed `Long64_t` (≤ 2^63−1) and 2^64 rows in a container isn't physically realizable, so the sentinel can't alias a real row.

If we go to `optional`, the clean version does it end-to-end — `fill()`/`fillContainer()` return `std::optional<std::uint64_t>` and `kInvalidRowId` disappears too — otherwise we'd have optional in `Token` and a sentinel one layer down, which is worse than either alone. That touches all the ROOT `fill()` overrides, so I'd rather not fold it into this PR. Happy to take it as a follow-up if that's the preference.

form/root_storage/root_rfield_write_container.cpp
pcanal: Indeed. The general condition 'something bad happened' between 	`fill()` and `commit()` is "real" and need to be decided upon (and then tested for).

wwuoneway: Thanks — and agreed the mechanism is exactly that: `fill()` returns the pending entry index and it isn't persisted until `commit()`, so an exception in between would leave a Token pointing at a row that never landed.

For this PR it's harmless in practice: the returned Token is discarded by the production write path today (nothing stores or navigates by it yet), so there's no live consequence. It starts to matter in the follow-up where FORM collects Tokens and passes them to `commitOutput()` — and that's the natural place to handle it, by only registering Tokens for entries that actually committed. We'll consider it in that follow-up. Good catch.

wwuoneway: @coderabbitai please create a tracking issue for this —  `write`/`commit` crash-consistency in the RNTuple path (`ROOT_RField_Write_ContainerImp::fill()` returns the pending row index, but an exception between `fill()` and `commit()` would leave a Token pointing at a row that was never persisted). To be addressed in the follow-up PR that starts consuming Tokens (only register Tokens for entries that actually committed). Please reference this PR/thread in the issue.

form/storage/storage_write_container.cpp
wwuoneway: Similar comment (copy reply below): 

Good point, and I agree std::optional is attractive for the type-safety (can't read the id without checking) more than the 2^64 case. One clarification: Token itself doesn't use a sentinel — it's uint64_t m_id + bool m_hasId, so there's no id-value collision risk in the token; the max() sentinel only lives in the write chain (fill()/fillContainer() → kInvalidRowId). And that one is unreachable in practice: TBranch::GetEntries() is a signed Long64_t (≤ 2^63−1) and 2^64 rows in a container isn't physically realizable, so the sentinel can't alias a real row.

If we go to optional, the clean version does it end-to-end — fill()/fillContainer() return std::optional<std::uint64_t> and kInvalidRowId disappears too — otherwise we'd have optional in Token and a sentinel one layer down, which is worse than either alone. That touches all the ROOT fill() overrides, so I'd rather not fold it into this PR. Happy to take it as a follow-up if that's the preference.

form/root_storage/root_rfield_write_container.cpp
gemmeren: I guess, that's where an 'isValid' flag in the token could help... But in practice I don't see a way the commit() would fail and the job wouldn't...

And as @wwuoneway points out, the tokens are 'tentative' , not yet being used, and planned to (future PR) be persistified within commit().

So, I agree, for now having invalid tokens (which we do not yet use) in case of a failed commit() is harmless.

form/core/token.hpp
wwuoneway: This ties into the discussion on the `id()`/`hasId()` accessors above (comment on same `std::optional` question. See [discussion](https://github.com/Framework-R-D/phlex/pull/830#discussion_r3823378959). I'd like to keep it out of this PR.

@coderabbitai please create a tracking issue: evaluate converting `Token`'s id representation to `std::optional<std::uint64_t>` (replacing `m_id` + `m_hasId` and the` id()`/`hasId()` accessors), and decide whether the write-chain sentinel `kInvalidRowId` in istorage.hpp is removed at the same time so the two layers stay consistent.  Please reference this PR/thread and note reviewer interest from the `std::optional` discussion.

form/root_storage/root_rfield_write_container.cpp
wwuoneway: Good instinct — my understanding is yes, this is accurate specifically because we use the single-threaded `ROOT::RNTupleWriter`, with one shared `RRawPtrWriteEntry` and a single `Fill()` per `commit()`. If we move to `RNTupleParallelWriter`m, this is a different story and we need revisit.
(@aolivier23 please correct me if my understanding is wrong. thanks)

@coderabbitai please create a tracking issue: `ROOT_RField_Write_ContainerImp::fill()` derives the pending row from `RNTupleWriter::GetNEntries()`, which is only correct under the single-threaded `ROOT::RNTupleWriter` we use today. If FORM moves to `RNTupleParallelWriter`, each thread fills an independent `RNTupleFillContext` and final row indices are only assigned at cluster flush/merge, so the pending row cannot be derived from `GetNEntries()` — the returned Token id would be wrong/racy. Revisit the row-index computation when/if parallel RNTuple writing is adopted. Please reference this PR/thread.

form/root_storage/root_tbranch_write_container.cpp
```diff
@@ -114,6 +114,10 @@ void ROOT_TBranch_Write_ContainerImp::fill(void const* data)
   }
   m_branch->Fill();
   m_branch->ResetAddress();
+
+  // 0-based entries: GetEntries() is the total count after this Fill(); row = count - 1.
+  // GetEntries() >= 1 here (we just filled), so the row is non-negative.
+  return static_cast<std::uint64_t>(m_branch->GetEntries() - 1);
```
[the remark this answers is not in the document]
wwuoneway: Good catch — these are two separate concerns. The L118-119 comment only justifies the cast (`GetEntries() >= 1` for non-negative row), it doesn't cover `Fill()` failing. You're right that `TBranch::Fill()` returns a negative byte count on a write error while fEntries has already been incremented, so we could return a Token for a row that wasn't persisted. **Here is the fix:** capture the return value, reset the address, and throw on a negative result before computing the row.  Done.

test/form/form_storage_test.cpp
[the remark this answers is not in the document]
wwuoneway: Done — wrapped both `readContainer` results in `std::unique_ptr<std::vector<int> const>` so the payloads are released at scope exit, which clears the leak noise from these allocations under Valgrind. Also added `<memory>` to the test includes so we're not relying on a transitive one.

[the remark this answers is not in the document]
aolivier23: Nailed it.  Thanks.

aolivier23 (dismissed):
LGTM

form/root_storage/root_rfield_write_container.cpp
aolivier23: I think we used to have a requirement that FORM always writes valid files even when a phlex job fails.

Agree this is fine for now.  I'll write an issue if I feel strongly about it.

form/root_storage/root_rfield_write_container.cpp
aolivier23: Pretty sure @pcanal is correct.  We're staying single-thread-only for this release of FORM.  Multi-thread support will have to address this.  I think I worked through a mental exercise once and concluded there's a way to get reasonable entry numbers with `RNTupleParallelWriter`.  So I think we're safe leaving this for later.

gemmeren: If you feel that there is a vulnerability, you could use a mutex to serialize the execution. [Only because Phlex runs modules concurrently, does not mean we have to use  `RNTupleParallelWriter` , but things have to be thread-safe.]

form/root_storage/root_tbranch_write_container.cpp
```diff
@@ -112,8 +112,18 @@ void ROOT_TBranch_Write_ContainerImp::fill(void const* data)
...
-  m_branch->Fill();
+  // TBranch::Fill() returns the number of bytes committed, or a negative value on a write error.
+  // ROOT increments entry count before a basket write can fail, so check return value first
+  Int_t const nbytes = m_branch->Fill();
   m_branch->ResetAddress();
+  if (nbytes < 0) {
+    throw std::runtime_error("ROOT_TBranch_Write_ContainerImp::fill TBranch::Fill() failed for " +
+                             col_name());
+  }
+
```
pcanal, line 123: This is needed and look correct but is not yet exercise in a test (see https://app.codecov.io/gh/Framework-R-D/phlex/pull/830?src=pr&el=tree&filepath=form%2Froot_storage%2Froot_tbranch_write_container.cpp&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Framework-R-D#diff-Zm9ybS9yb290X3N0b3JhZ2Uvcm9vdF90YnJhbmNoX3dyaXRlX2NvbnRhaW5lci5jcHA=)

form/persistence/persistence_writer.cpp
```diff
@@ -48,13 +48,17 @@ void PersistenceWriter::createContainers(
...
+Token PersistenceWriter::registerWrite(std::string const& creator,
+                                       std::string const& label,
+                                       void const* data,
+                                       std::type_info const& type)
 {
   std::unique_ptr<Placement> plcmnt = getPlacement(creator, label);
-  m_store_writer->fillContainer(*plcmnt, data, type);
+  std::uint64_t const row = m_store_writer->fillContainer(*plcmnt, data, type);
+  if (row == kInvalidRowId) {
+    return Token{plcmnt->fileName(), plcmnt->containerName(), plcmnt->technology()};
```
gemmeren: This seems to be a failure condition, with invalid row number, FORM won't be able to find the data.
Should we throw here?

gemmeren (dismissed):
Thanks @wwuoneway 
I leave it up to you whether you want to use optional (which seems like a good strategy).
Other than that, this looks good to me.

form/root_storage/root_rfield_write_container.cpp
aolivier23: We should be fine for now.  I think the whole file is inside a mutex right now.

aolivier23: Also, while it's not a testing-strength guarantee, I've been running a lot of 4-thread phlex jobs that write out using RNTuple lately.  I'm not encountering any obvious runtime problems.

form/root_storage/root_tbranch_write_container.cpp
wwuoneway: Done — Added a test ("Root TBranch fill: throws when `TBranch::Fill()` reports a write error") that now exercises this path. It wires a real `TTree`/`TBranch` through the factories, shrinks the branch basket so a few fills force a basket flush, then marks the `TFile` non-writable so the flush fails — `TBranch::Fill()` returns a negative value and `fill()` throws. Thanks.

form/persistence/persistence_writer.cpp
wwuoneway: `kInvalidRowId` here isn't a failure — it's the documented "no addressable rows" return (see the `fill()`/`fillContainer()` contract in `istorage.hpp`). It comes only from the generic/no-op backend (`Storage_Write_Container::fill`). Genuine write errors already throw inside the backend at the point of failure (e.g. the `TBranch::Fill()` guard, see above comment for test coverage), so by the time we're back here the sentinel can only mean "this technology doesn't address rows," never "the write failed." That's exactly what the not-set Token (`hasId() == false`) represents, so throwing here would break the generic backend and conflate the two cases. Consumers guard with `hasId()` before calling `id()`. I've added a comment to make the branch's intent explicit.

wwuoneway:
> Thanks @wwuoneway I leave it up to you whether you want to use optional (which seems like a good strategy). Other than that, this looks good to me.

Agreed `std::optional<std::uint64_t>` is the cleaner representation. Since it changes `Token`'s public accessors and the write-chain sentinel — and overlaps with the pending naming/convention pass over `form/` and `test/form/` — I'd like to keep this PR scoped to just returning the `Token` and do the `optional` refactor as a focused follow-up. The current sentinel + `hasId()` design is correct and tested, so nothing is blocked by deferring.
I've already opened a tracking issue for this #833.

form/persistence/persistence_writer.cpp
gemmeren: So how would we locate the written data object on read if we don't have a valid row number? If we can't, then this _is_ a failure. (even though the data has been written)

aolivier23 (dismissed):
LGTM

form/root_storage/root_rfield_write_container.cpp
pcanal: >  I think the whole file is inside a mutex right now.

Is a global mutex or a per `TFile` mutex?

form/persistence/persistence_writer.cpp
wwuoneway: Good point and you're right — once these Tokens become the read-side navigation keys (follow-up PR), a not-set Token is a locator to nothing, so a product routed there is never locatable regardless of row on read. I've changed this and throw here. I've also updated the inline code comment and test to reflect this. Thanks.

form/root_storage/root_rfield_write_container.cpp
wwuoneway: My understanding is that the `GetNEntries()` row is accurate as long as a given `RNTuple` is never written from multiple threads at once — and today that holds, because writes to a given output file/RNTuple are serialized by the phlex output node (registered `concurrency::serial`, effectively per-TFile), not by anything in FORM itself. So the single-threaded-per-RNTuple assumption is currently satisfied. If this is later changed, it will need revisiting as mentioned in #834.

gemmeren: This is independent of RNTupleParallelWriter , I would encourage to have FORM be threadsafe (e.g. its own mutex) w/o necessarily being concurrent and not reliant on Phlex. This should be rather straightforward, but could be a separate PR (there may be other points too, where mutex may be needed).

wwuoneway: Agreed, and good point. You're right that this is independent of `RNTupleParallelWriter` — FORM's write side currently has no internal mutex and leans on the Phlex serial output node, which it shouldn't bank on but a natural, self-contained case. Best as a separate PR as you suggest.

wwuoneway: @coderabbitai please open a tracking issue:
**Title:** Make FORM's write path thread-safe independent of Phlex

**Body:**
FORM's write path currently has no internal locking and relies on Phlex registering the output node with `concurrency::serial` for serialization. FORM should be thread-safe on its own — not concurrent by default, but safe without depending on Phlex's scheduling.

Possible solutions: Add FORM-owned write-path mutex(es), mirroring the existing read-side pattern (root_tbranch_read_mutex, root_rfield_read_mutex). Audit for other call sites that may need a mutex.
Independent of #834  (RNTupleParallelWriter). Please reference this PR/thread.

aolivier23 (approved):
LGTM

gemmeren (approved):

PR #837, opened 2026-08-21
fix: resolve dangling clang-tidy warnings in core and tests

knoepfel:
These are straggling `clang-tidy` errors that should have been resolved earlier.

- **Code quality**
  - Resolved remaining `clang-tidy` warnings across core, storage, plugin, and test code.
  - Replaced manual maximum updates with `std::max` in `index_router.cpp`.
  - Improved type-safe comparisons with `std::cmp_equal` and `std::cmp_greater_equal`.
  - Simplified destructor, mutex guard, pointer declaration, namespace, and redundant return syntax.
  - Updated `layer_generator::execute` to pass `data_cell_index_ptr` by value and documented coroutine lifetime ownership.
  - Updated `WaveformGenerator` to use an inline defaulted destructor.

- **Tests**
  - Updated `flush_gate` assertions to compare signed and unsigned counts safely.

- **Documentation**
  - Marked `bugprone-throw-keyword-missing` as fixed in the `clang-tidy` report.
  - Added the associated change reference.


knoepfel:
@coderabbit review

marcpaterno (approved):

PR #842, opened 2026-08-21
test: standardize C++ test source and target names

knoepfel:
This PR does not change any functionality.  The purpose of this PR is to:

1. Name all unit-test source files with a trailing suffix `<name>_test.cpp`.
2. Name all unit tests `<name>` where possible, omitting the `_test` suffix as it is not necessary.

This avoids situations where the implmentation of one facility might be located at `phlex/core/facility.cpp` and the test for that facility might be at `test/facility.cpp`.  Although the paths of those files unambiguously distinguish between the files, it can be difficult in an editor to determine at a glance which files are which.

- This PR does not add, remove, or change test coverage.
- This PR does not change production APIs or functional behavior.
- It only standardizes C++ test source names, CMake registrations, and test target names.


knoepfel:
@coderabbit, your section on "Test coverage" in the body of the PR is baffling—there should be no changes in test coverage at all.

knoepfel:
@coderabbit, can you now fix the body of the PR, as described at https://github.com/Framework-R-D/phlex/pull/842#issuecomment-5375647498.

marcpaterno (approved):

PR #850, opened 2026-08-26
feat: add resource-limited execution to graph nodes

### LINKED ISSUE #851: Phase 1 resource support
## Status

Implemented by [PR #850](https://github.com/Framework-R-D/phlex/pull/850); pending merge.

## Scope

Phase 1 supports resources registered directly on `framework_graph` by code compiled against the concrete resource type. It provides two resource categories:

1.  An unlimited-access resource is a graph-owned, non-data process facility with no `token_type`. It imposes no resource-capacity limit and may therefore be used concurrently by any number of algorithm invocations, subject to the facility's own thread-safety contract.
2.  A serialized resource has a `token_type`, but no `tokens()`, and permits one algorithm invocation at a time.

A declaration with both `token_type` and `tokens()` is recognized as a multi-token resource form, but Phase 1 rejects it at compile time. It must not be silently treated as a serialized resource.

For bounded resources, the access value is a concurrency token held for the complete execution of the algorithm body. For unlimited resources, the access value grants access only and does not reserve capacity.

## Unlimited-Access Resources

An unlimited-access resource is a graph-owned, process-lifetime facility that algorithms may use concurrently. It is not a data product: it has no data-layer association, producer provenance, output representation, or persistence semantics.

Use an unlimited-access resource for a non-data capability such as a counter-based random-number generator, thread-safe service client, library context, logging or metrics endpoint, device-runtime interface, or implementation cache. The resource author does not declare `token_type` or `tokens()`.

```cpp
#include <Random123/philox.h>

g.add_unlimited_resource<r123::Philox4x32>();
```

An algorithm requesting the resource receives a pointer to the graph-owned facility. By default, the pointer is `T const*`, so the facility must support concurrent read operations:

```cpp
belled_whistle add_bell(int whistle_number, whistle const& w, r123::Philox4x32 const rng)
{
  auto key = r123::Philox4x32::key_type{/* build out of whistle_number */};
  auto counter = r123::Philox4x32::ctr_type{...};
  return belled_whistle{w, rng(counter, key)};
}

m.transform("whistle_decorator", 
            [](handle<whistle> w, r123::Philox4x32 const* rng) {
              return add_bell(w.id_number(), *w, *rng);
            },
            concurrency::unlimited)
  .input_family(selector, resource<philox_rng>{});
```

This example shows creating anew the key and counter values for each invocation of the `add_bell` algorithm.

The resource object address remains stable for the lifetime of the `framework_graph`. At the catalog boundary, `resource_catalog::access_for<T>()` returns `gsl::not_null<T const*>`. The node-level cache and algorithm-facing signature use ordinary `T const*` values. No TBB resource limiter is created, so a node using only unlimited-access resources remains an ordinary `function_node` or `multifunction_node` and retains its configured Phlex concurrency.

An unlimited-access resource is not the appropriate abstraction merely because an object is read-only and shared for a whole job. If an object represents scientific or reproducible state whose identity, origin, version, or contents may need provenance or persistence, it should be a data product at the appropriate layer.

`const` expresses mutation rights, not resource concurrency policy. Mutable concurrent access requires an explicit future access-policy spelling.

## Serialized Resources

A serialized resource declares an algorithm-facing token type that the framework can construct from its owned resource object. Phase 1 supports either a pointer-compatible token or a copyable value token.

A pointer-compatible resource-object token can be `T*` or `T const*`:

```cpp
class database {
public:
  using token_type = database const*;

  explicit database(connection_options options);
};

g.add_serialized_resource<database>(options);
```

```cpp
void query(request const& request, database const* database_resource);

m.observe("query", query, concurrency::unlimited)
  .input_family(selector, resource<database>{});
```

The framework creates one limiter token from the address of its framework-owned `database` object and holds it while `query` executes. A pointer-compatible token refers to that framework-owned object.

A serialized resource may instead produce a stateless, copyable token solely to regulate access to a thread-unsafe library:

```cpp
struct thread_unsafe_library {
  struct access_token {};
  using token_type = access_token;

  static token_type token() { return {}; }
};

g.add_serialized_resource<thread_unsafe_library>();
```

```cpp
void call_library(request const& request, thread_unsafe_library::access_token);

m.observe("call_library", call_library, concurrency::unlimited)
  .input_family(selector, resource<thread_unsafe_library>{});
```

The framework creates and stores one `access_token` in the serialized-resource limiter. An algorithm receives that token by value for the complete invocation, ensuring that no two calls which depend on this resource enter the thread-unsafe library concurrently. The token has no state and does not provide access to a catalog-owned resource object.

The effective maximum concurrency for a node is the minimum of the operator's concurrency setting and the resource's serialized-resource capacity. A distinct, copyable token type may instead be produced by a `token()` member on the registered resource.  A token type that cannot be produced from `T*`, is not exactly `T`, and has no `token()` member is not a Phase 1 resource declaration.

## Declaring Dependencies

All resource categories use the same registration-time marker:

```cpp
resource<T>{}
```

Resources follow product selectors in `input_family(...)` and correspond, in order, to trailing algorithm parameters:

```cpp
void process(input const&, geometry const*, database const*);

m.observe("process", process, concurrency::unlimited)
  .input_family(selector, resource<geometry>{}, resource<database>{});
```

The registration layer counts resource markers as callable inputs but uses only preceding product-selector parameters for product type deduction. It does not assume resource parameters are pointers, so value-token parameters are valid.

Resource types identify registered resources. A graph permits one registration of each concrete resource type, and a dependency list may name each resource type at most once; duplicate resource dependencies are rejected at compile time.

## Framework Responsibilities

`framework_graph` owns registered resource objects and their limiter state. Both must outlive graph nodes, since limited nodes retain references to their TBB resource limiters.

Resource registration must occur before algorithm registration references the resource. Referencing an unregistered resource is a registration error with a clear diagnostic.

For a node that uses multiple limited resources, the framework acquires one token from each resource for the whole body invocation.

## Exclusions

Phase 1 does not enable multi-token resources; multi-token resources are covered by [Phase 3 (#853)](https://github.com/Framework-R-D/phlex/issues/853).

knoepfel:
## Summary of changes

Adds resource-limited graph-node execution. Users can declare resources at graph construction, request typed resource tokens alongside product selectors, and bound a resource's concurrent use. The implementation supplies a resource catalog and registration API, separates product and resource dependencies, selects the appropriate TBB node form, and extends observer, transform, predicate, fold, and unfold nodes. It also adds focused API, node-builder, and concurrency-limit coverage, plus the required TBB preview-feature configuration.

## Recommended approach for reviewing

Review the commits in order; each is independently buildable and narrows the next commit's context:

1. **TBB capability check** (`171c8271`): verify the preview feature is detected and enabled only where needed.
2. **Resource model and registration** (`94231402`): review the public resource concepts, catalog ownership, lookup/error behavior, tokens, and dependency classification; use `resources_test.cpp` as the behavioral guide.
3. **Registration boundary** (`9afd615a`): confirm resources are partitioned cleanly from product selectors without regressing existing registration calls.
4. **Execution integration** (`e0f69d09`): focus on `node_builder.hpp`: resource argument ordering, bounded-resource detection, and the selected TBB node type. `node_builder_test.cpp` documents the intended mechanics.
5. **Node-family adoption** (`2d9d2dc4` through `47b8e5a0`): these follow the same integration pattern. Review one representative node carefully, then inspect the others primarily for consistency and their focused tests.

The final test changes demonstrate both token delivery and that a resource shared by nodes enforces the configured concurrency bound.


test/fold_test.cpp
```diff
@@ -127,3 +134,42 @@ TEST_CASE("Fold output without send consumed downstream", "[graph]")
...
+
+  auto counter = detail::thread_counter::counter_type{};
+  auto add_with_resource = [&counter](std::atomic<unsigned int>& sum,
+                                      unsigned int number,
+                                      fold_resource_1 const*,
+                                      fold_resource_2 const*) {
+    detail::thread_counter throw_if_more_than_one_thread{counter};
+    detail::sleep_for(5ms);
+    sum += number;
+  };
```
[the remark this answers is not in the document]
knoepfel: @coderabbit, there is already another `unfold` test that checks load-bearing tokens.  The purpose of this test is to check that multiple resources can serve as inputs to a fold operator.

phlex/core/declared_observer.hpp
```diff
@@ -44,11 +46,15 @@ namespace phlex::detail {
...
   class observer_node : public declared_observer {
-    using input_args = AlgorithmBits::input_parameter_types;
     using function_t = AlgorithmBits::bound_type;
-    static constexpr auto num_inputs = AlgorithmBits::number_inputs;
+    static constexpr auto num_resources = sizeof...(Resources);
+    static constexpr auto num_products = AlgorithmBits::number_inputs - num_resources;
+    using input_product_types = AlgorithmBits::template input_parameters<num_products>;
+    using builder =
+      node_builder<messages_t<num_products>, function_t, no_outputs_t, std::tuple<Resources...>>;
+    using node_t = typename builder::node_t;
```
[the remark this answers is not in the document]
knoepfel: @coderabbit, look at the most recent (forced) push.  There is now an explicit `product_selector` first argument.

phlex/core/resource_api.hpp
```diff
@@ -0,0 +1,149 @@
...
+    // std::tuple<resource<T>...> for the resources that follow the split.
+    using resources_type = boost::mp11::mp_rename<resource_types, std::tuple>;
+
+    static constexpr std::size_t number_selectors = index;
+    static constexpr std::size_t number_resources = boost::mp11::mp_size<resource_types>::value;
+
+    // Once the first resource has been seen, every remaining argument must also be a resource;
+    // otherwise the arguments cannot be partitioned into a selector group and a resource group.
+    static constexpr bool resources_are_last =
+      boost::mp11::mp_all_of<resource_types, internal::is_resource_p>::value;
```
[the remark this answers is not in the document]
knoepfel: @CodeRabbit, look at the most recent (forced) push.

test/resources_test.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include "phlex/utilities/thread_counter.hpp"
+
+#include "catch2/catch_test_macros.hpp"
+#include "catch2/matchers/catch_matchers_string.hpp"
+
+using namespace phlex;
+using namespace phlex::detail;
+
+namespace {
+  struct Catch2 {};
```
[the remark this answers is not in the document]
knoepfel: @CodeRabbit, look at the most recent (forced) push.  I decided to suppress this warning.

test/resources_test.cpp
```diff
@@ -0,0 +1,67 @@
...
+  g.provide(
+     "number_maker", [](data_cell_index const&) { return 42; }, concurrency::unlimited)
+    .output_product("number_maker", "", "job");
+
+  auto counter = thread_counter::counter_type{};
+  auto verify_number = [&counter](int const num, Catch2 const*) {
+    // Both observers share one resource token and must not run concurrently.
+    thread_counter throw_if_more_than_one_thread{counter};
+    REQUIRE(num == 42);
+  };
```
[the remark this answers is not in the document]
knoepfel: @CodeRabbit, look at the most recent (forced) push.

phlex/core/declared_observer.hpp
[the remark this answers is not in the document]
knoepfel: @coderabbit, see the most recent forced push.

test/tbb-preview/CMakeLists.txt
```diff
@@ -1,52 +1,11 @@
...
-    LIBRARIES
-    phlex::core
-    TBB::tbb
-    spdlog::spdlog
-  )
-  target_compile_definitions(resource_limiting PRIVATE TBB_PREVIEW_FLOW_GRAPH_RESOURCE_LIMITING=1)
-else()
-  message(STATUS "Skipping resource_limiting test: flow::resource_limiter not available in TBB")
-endif()
+target_compile_definitions(resource_limiting PRIVATE TBB_PREVIEW_FLOW_GRAPH_RESOURCE_LIMITING=1)
```
knoepfel, line 11: Check whether this can be removed.

test/resources_test.cpp
marcpaterno, line 1: We should create an issue to provide guidance for the organization of tests, with an eye toward making them easier for human readers to understand.

The specific idea in this case is:
1. move types used only for testing to a headers that can be shared between tests
2. make the individual test files for narrowly focused (and thus smaller)

phlex/core/resource/concepts.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+  template <typename T, typename... Args>
+  concept unlimited_resource_registration =
+    unlimited_resource<T> && !std::is_const_v<T> && std::constructible_from<T, Args...>;
+
+  template <typename T, typename... Args>
+  concept serialized_resource_registration =
+    serialized_resource<T> && !std::is_const_v<T> && std::constructible_from<T, Args...>;
+
+  namespace internal {
```
marcpaterno, line 44: This is hard to understand. Consider instead:

1. general template not defined
2. specialization for each of serialized and unlimited
3. pool added later, when it is supported.

phlex/core/resource/catalog.hpp
```diff
@@ -0,0 +1,80 @@
...
+      using resource_type = Entry::resource_type;
+      auto const found = resources_.find(std::type_index(typeid(resource_type)));
+      if (found == resources_.end()) {
+        throw std::runtime_error(fmt::format("Resource of type '{}' has not been registered",
+                                             boost::core::demangle(typeid(resource_type).name())));
+      }
+      return dynamic_cast<Entry*>(found->second.get());
+    }
+
+    std::map<std::type_index, std::unique_ptr<resource_base>> resources_;
```
marcpaterno, line 76: Consider using std::unordered_map

phlex/core/resource/catalog.hpp
```diff
@@ -0,0 +1,80 @@
...
+    auto& limiter_for() const
+    {
+      auto* entry = entry_for<serialized_resource_entry<T>>();
+      assert(entry != nullptr && "resource catalog entry has unexpected serialized entry type");
+      return entry->limiter();
+    }
+
+  private:
+    template <typename Entry>
+    Entry* entry_for() const
```
marcpaterno: Consider changing the return type to `gsl::not_null<Entry>` so that the type makes clear that by construction of the whole system, the pointer can not be null (the dynamic_cast can not fail because constraint checking has ensured it).

test/resources_test.cpp
knoepfel: @coderabbit, can you make an issue describing @marcpaterno's idea?

test/tbb-preview/CMakeLists.txt
knoepfel: Removed by most recent push.

phlex/core/resource/concepts.hpp
knoepfel: See commit https://github.com/Framework-R-D/phlex/pull/850/commits/d9ee74e0565b3dd8dbebd5c9ad608ddedb3530fa.

phlex/core/resource/catalog.hpp
knoepfel: See commit https://github.com/Framework-R-D/phlex/pull/850/commits/fdd0700356a81e8672e30379f49ab9463ca4763c.

phlex/core/graph_proxy.hpp
```diff
@@ -191,7 +195,7 @@ namespace phlex::detail {
                               std::string destination_data_layer,
                               concurrency c) const
   {
-    return glue<Splitter>{graph_, nodes_, nullptr, errors_, config_}.unfold(
+    return glue<Splitter>{graph_, nodes_, nullptr, errors_, resources_, config_}.unfold(
```
marcpaterno: Consider a new issue to rename Splitter to reflect newer naming (perhaps Unfolder)

phlex/core/resource_api.hpp
```diff
@@ -0,0 +1,71 @@
...
+namespace phlex::detail {
+  namespace internal {
+    template <typename T>
+    struct is_resource_impl : std::false_type {};
+
+    template <typename T>
+    struct is_resource_impl<resource<T>> : std::true_type {};
+
+    template <typename T>
+    using is_resource_p = is_resource_impl<std::remove_cvref_t<T>>;
```
marcpaterno: Consider a more explicit name: is_resource_predicate

phlex/core/registration_api.hpp
```diff
@@ -14,28 +14,49 @@
...
 
 namespace phlex {
   class configuration;
 }
 
 namespace phlex::detail {
 
+  template <typename T>
+  concept selector_or_resource =
+    std::same_as<std::remove_cvref_t<T>, product_selector> || is_resource<T>;
```
marcpaterno: Consider adding is_product_selector as a concept, so that this line is simpler (especially if it would also be used anywhere else to simplify code).

marcpaterno (commented):
This is not yet a complete review.

phlex/core/registration_api.hpp
knoepfel: Introduced in commit https://github.com/Framework-R-D/phlex/pull/850/commits/9b99a1ffe0cc3aa521836354792fe8d0ca8e631b.

phlex/core/graph_proxy.hpp
knoepfel: Since the required change is local only to the `graph_proxy.hpp` file, I made the change in commit https://github.com/Framework-R-D/phlex/pull/850/commits/fdeb1ee7ed23a064b56c472136a43b10d2d71737.

phlex/core/resource/catalog.hpp
knoepfel: See commit https://github.com/Framework-R-D/phlex/pull/850/commits/2e003c15206e6da331f08f176d8b03ab209b8881

phlex/core/node_builder.hpp
```diff
@@ -0,0 +1,291 @@
...
+        std::forward<decltype(arguments)>(arguments));
+    }(std::index_sequence_for<Resources...>{});
+  }
+
+  template <typename InputMessages, typename Function, typename... Resources>
+    requires(not has_serialized<Resources...>)
+  struct node_builder<InputMessages, Function, no_outputs_t, std::tuple<Resources...>> {
+    using node_t = node_type_for_t<InputMessages, no_outputs_t, std::tuple<Resources...>>;
+
+    template <typename NodeBody>
```
marcpaterno: Consider adding some comments. Perhaps:

1. A "preamble" noting the 6 different 'make_nodes', and
2. A comment associated with each, specifying what kind of nodes it is for.

For example, the first make_node is for making Observers...

phlex/core/node_builder.hpp
```diff
@@ -0,0 +1,291 @@
...
+                       Function ft,
+                       NodeBody node_body)
+    {
+      auto unlimited = resource_dependencies<Resources...>::unlimited_resource_accesses(resources);
+      return {
+        g,
+        concurrency,
+        resource_dependencies<Resources...>::serialized_resource_limiters(resources),
+        [ft = std::move(ft), node_body = std::move(node_body), unlimited = std::move(unlimited)](
+          InputMessages const& messages, auto&, auto&... resource_tokens) mutable {
```
marcpaterno: This would be clearer if the unnamed `auto&` argument were given a /* what i am */ comment inline.

phlex/core/node_builder.hpp
```diff
@@ -0,0 +1,291 @@
...
+          return std::invoke(callable,
+                             std::forward<FunctionArguments>(function_arguments)...,
+                             resource_argument<Resources, resource_local_indices<Resources...>[Is]>(
+                               unlimited_resources, serialized_tokens)...);
+        },
+        std::forward<decltype(arguments)>(arguments));
+    }(std::index_sequence_for<Resources...>{});
+  }
+
+  template <typename InputMessages, typename Function, typename... Resources>
```
marcpaterno: Consider removing the template parameter Function from the struct template node_builder, and moving it to the smaller scope of the `make` static member function template.

knoepfel: See commit https://github.com/Framework-R-D/phlex/pull/850/commits/fd1d6355e9b39bb8ba59104af32174cec9a1f4aa.

phlex/core/node_builder.hpp
knoepfel: See commit https://github.com/Framework-R-D/phlex/pull/850/commits/1eb430c862736fc790badc9fb8386d60f3a3a2bd.

phlex/core/node_builder.hpp
knoepfel: See commit https://github.com/Framework-R-D/phlex/pull/850/commits/1eb430c862736fc790badc9fb8386d60f3a3a2bd.

test/node_builder_test.cpp
```diff
@@ -0,0 +1,310 @@
...
+  resources.add_serialized<value_token_resource_type>(42);
+  std::atomic<bool> called{};
+
+  using builder_t = node_builder<int, no_outputs_t, std::tuple<value_token_resource_type>>;
+  auto node = builder_t::make(
+    graph,
+    concurrency::serial.value,
+    resources,
+    [&called](int value, value_token_resource_type token) {
+      CHECK(value == 42);
```
marcpaterno: Consider changing two of the 42s to a different number, so that is is clear which value in the test is the same as which value in the setup.

test/node_builder_test.cpp
```diff
@@ -0,0 +1,310 @@
...
+  resources.add_serialized<custom_token_resource_type>();
+  std::atomic<bool> called{};
+
+  using builder_t = node_builder<int, no_outputs_t, std::tuple<custom_token_resource_type>>;
+  auto node = builder_t::make(
+    graph,
+    concurrency::serial.value,
+    resources,
+    [&called](int value, int token) {
+      CHECK(value == 42);
```
marcpaterno: Same comment as above.

test/node_builder_test.cpp
knoepfel: See commit https://github.com/Framework-R-D/phlex/pull/850/commits/24ae7545ceb04f43a7281edf4f138ad06d1c0e1e.

test/node_builder_test.cpp
knoepfel: See commit https://github.com/Framework-R-D/phlex/pull/850/commits/24ae7545ceb04f43a7281edf4f138ad06d1c0e1e.

marcpaterno (approved):

PR #861, opened 2026-09-01
refactor(python): split configuration and registration wrappers

knoepfel:
This PR is a step toward addressing the `readability-function-size` clang-tidy warning.

- **Code quality**
  - Refactored Python configuration and registration wrappers into smaller helper functions.
  - Reduced function size to address `readability-function-size` warnings.
  - Added a lint suppression for the anonymous `get_ffi_type` function.

- **Configuration wrapper**
  - Extracted scalar, array, string-map, and vector-to-tuple conversions into dedicated helpers.
  - Improved cached-value handling by incrementing the reference count before returning cached objects.
  - Added UTF-8 decode failure checks for configuration keys.
  - Added the required `<map>` and `<vector>` includes.

- **Registration wrapper**
  - Split transform registration into helpers for function-pointer extraction, output validation, selector construction, type restrictions, and callback registration.
  - Added structured provider-registration argument parsing and validation.
  - Added managed-reference support for wrapped callables.
  - Separated scalar and collection output registration paths.

### DIFF
--- a/plugins/python/src/configwrap.cpp
+++ b/plugins/python/src/configwrap.cpp
@@ -83,7 +83,7 @@ static PyObject* string_map_to_python(std::map<std::string, std::string> const& 
 }
 
 template <typename T, typename Converter>
-static PyObject* vector_to_python_tuple(std::vector<T> const& values, Converter&& convert)
+static PyObject* vector_to_python_tuple(std::vector<T> const& values, Converter const& convert)
 {
   auto const size = checked_tuple_size(values.size());
   if (!size) {
--- a/plugins/python/src/configwrap.cpp
+++ b/plugins/python/src/configwrap.cpp
@@ -1,7 +1,9 @@
 #include <cstdint>
+#include <map>
 #include <optional>
 #include <ranges>
 #include <string>
+#include <vector>
 
 #include "phlex/configuration.hpp"
 #include "wrap.hpp"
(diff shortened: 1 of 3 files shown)


wlav:
Note that these changes affect the updates made for #817 . Fine per se, but it means that those updates can no longer be directly applied after #817 goes in, so plan for some time to update.

plugins/python/src/configwrap.cpp
```diff
@@ -56,6 +58,126 @@ static std::optional<Py_ssize_t> checked_tuple_size(std::size_t n)
...
+  if (!result) {
+    return nullptr;
+  }
+  for (Py_ssize_t i = 0; i < *size; ++i) {
+    PyObject* item = convert(values[static_cast<std::size_t>(i)]);
+    if (!item) {
+      Py_DECREF(result);
+      return nullptr;
+    }
+    if (PyTuple_SetItem(result, i, item) < 0) {
```
wlav: `PyTuple_SetItem` has zero chance of failing, given the earlier setup and checks, but coverage will ding you for it. It's why it wasn't checked before. In fact, this is fine place for `PyTuple_SET_ITEM` (the macro).

plugins/python/src/configwrap.cpp
```diff
@@ -56,6 +58,126 @@ static std::optional<Py_ssize_t> checked_tuple_size(std::size_t n)
...
+    return nullptr;
+  }
+
+  PyObject* result = PyTuple_New(*size);
+  if (!result) {
+    return nullptr;
+  }
+  for (Py_ssize_t i = 0; i < *size; ++i) {
+    PyObject* item = convert(values[static_cast<std::size_t>(i)]);
+    if (!item) {
```
wlav, line 100: Can only fail on out-of-memory.

plugins/python/src/modulewrap.cpp
```diff
@@ -1061,126 +1189,36 @@ static PyObject* md_transform(py_phlex_module* mod, PyObject* args, PyObject* kw
...
-    } else {
-      py_callback<dcarg, n> cb{callable};
-      insert_tranform_for_callback(cb);
-    }
-  };
-
-  if (!unroll_switch<max_supported_args>(input_selectors.size(), transform_n_args)) {
-    PyErr_SetString(PyExc_TypeError, "unsupported number of inputs");
+  if (!validate_numba_transform_types(cname, input_types, out_type, ccallf) ||
+      !register_transform_callback(
```
wlav: Not a fan of this one: the way this reads is that this fails if "numba transform types" fail to "validate", but underlying is a twist: if `ccallf` is `nullptr`, it succeeds ... because it's not a numba callback, hence nothing to "validate". It's probably more readable to have `!(ccallf && validate...)` and not have the `ccallf` check in `validate`. Or maybe precede with `bool isnumba = (bool)ccallf;`?

wlav (changes requested):
Nothing drastic. The follow-up on #817 does have a new feature (which allowed me to write easier tests), that doesn't have a good location anymore and may require a bit more refactoring when that lands.

plugins/python/src/configwrap.cpp
knoepfel: Comment added to this effect.

plugins/python/src/modulewrap.cpp
```diff
@@ -1061,126 +1189,36 @@ static PyObject* md_transform(py_phlex_module* mod, PyObject* args, PyObject* kw
...
   }
 
   // insert output converter node into the graph
-  auto out_pq = product_selector{.creator = identifier(pyname),
-                                 .layer = identifier(output_layer),
-                                 .suffix = identifier(pyoutput)};
+  std::string const pyname = "py_" + cname;
+  std::string const pyoutput = output_suffixes[0] + "_py";
+  auto out_pq = product_selector{
+    .creator = identifier(pyname), .layer = *output_layer, .suffix = identifier(pyoutput)};
```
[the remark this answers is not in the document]
knoepfel: @coderabbit, verify that your concern is addressed with the latest commit.

plugins/python/src/configwrap.cpp
knoepfel: `PyTuple_SET_ITEM` used.

plugins/python/src/modulewrap.cpp
knoepfel: Adopted something similar to what you suggested.

wlav (approved):

PR #862, opened 2026-09-01
Use struct for class with only public members

pcanal:
Following https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#rc-struct, a few `class`es should be marked as `struct`.  

`root_rntuple_write_container_imp` is reverted to be a `class`.  Eventhough `root_rntuple_write_container_imp` locally only has public members (including data members), it also has, via inheritance, private members.

- **Code quality**
  - Converted selected `class` declarations to `struct` where members are public.
  - Updated type traits in `phlex/metaprogramming/type_deduction.hpp` and `phlex/model/type_id.hpp`.
  - Updated `product_selectors_type_setter` in `phlex/core/product_selector.hpp`.

- **Access control**
  - Kept `root_rntuple_write_container_imp` as a `class` because it inherits private members.
  - Updated both its forward declaration and definition in `form/root_storage/root_rntuple_write_container.hpp`.

- **Tests and mocks**
  - Converted the mock `algorithm` template in `test/mock-workflow/algorithm.hpp`.
  - Converted `signed_value_provider` in `test/plugins/ij_source.cpp`.

- **Behavior**
  - Preserved public inheritance and member access.
  - Made no functional or API behavior changes.

### DIFF
--- a/form/root_storage/root_rntuple_write_container.hpp
+++ b/form/root_storage/root_rntuple_write_container.hpp
@@ -44,9 +44,9 @@ namespace form::detail::experimental {
 #endif
 
   class root_rntuple_write_container_imp : public storage_write_association {
+  public:
     root_rntuple_write_container_imp(std::string const& name);
     ~root_rntuple_write_container_imp() override;
-
     //Rule of five
     root_rntuple_write_container_imp(root_rntuple_write_container_imp const& other) = delete;
     root_rntuple_write_container_imp(root_rntuple_write_container_imp&& other) = delete;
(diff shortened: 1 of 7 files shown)


form/root_storage/root_rntuple_write_container.hpp
pcanal, line 66: @aolivier23 @wwuoneway Could you extend the documentation with a small rational why this are public even-though the inherited member `storage_write_container::ifile_` is private.  (And it feels that there might be so invariant/relations between those 3 ... i.e. the 'user' might not be able to assign to them arbitrarily)

aolivier23: Good catch.  We created issue #867  to make these data members private and provide them through accessors.

wwuoneway: +1, make sense. #867 give us the place to enforce that. Thanks!

knoepfel (approved):

PR #864, opened 2026-09-01
FORM write path: config-once, create container once, persistence owns the index

wwuoneway:
**Summary:**
Reworks the FORM write path so **FORM owns the product configuration and persistence is a stateless executor**. FORM parses config once, resolves each product to its destination placements, creates containers once, routes writes, and commits each place once per record. Persistence turns fully-resolved placements into storage calls, enforces the write-token invariant, and owns the navigation ("index") container end to end. On-disk output is unchanged.

**What changed:**
- **Configuration ownership → FORM** — config is parsed once at construction; a per-creator write plan is built lazily on first write (config-once, create-once). Persistence no longer holds product config.
- **Per-creator write plan + multi-destination** — a creator's plan resolves each product to all configured destinations, so one product can fan out to several files/technologies; a creator can also add products on a later write.
- Sh**ared naming vocabulary** — `build_full_label(creator, label)` moved from persistence into `core/placement`, so writer and reader share the `creator/label` convention without either owning the other's vocabulary.
- **Navigation index owned by persistence** — FORM names only product containers and commits each place with its segment id. Persistence derives each place's index placement from the product placement (resolved once at create, cached), creates the index container itself, records the id on commit, then finalizes the row. Interface: `commit_output` → `commit_place(placement, id)`; the transient `fill_index` seam was removed. FORM is now opaque to whether a place is indexed.

- **Code**
  - Moved `build_full_label` to `core/placement` for shared reader and writer use.
  - Moved product configuration parsing and write-plan ownership into FORM.
  - Added per-creator lazy write plans with product fan-out to multiple destinations.
  - Created containers once and committed each destination once per record.
  - Skipped unconfigured products without aborting the write.
  - Refactored persistence to accept resolved `placement` values.
  - Removed persistence configuration state and the `fill_index` seam.
  - Moved navigation index creation, ID recording, and row finalization into persistence.
  - Replaced `commit_output` with `commit_place`.
  - Added injectable persistence and storage-writer constructors for testing.

- **Tests**
  - Added tests for container creation across events.
  - Added tests for multi-destination product writes.
  - Added tests for unconfigured product handling.
  - Added coverage for non-row-addressed backend rejection.
  - Migrated persistence tests to the placement-based API.

### DIFF
--- a/form/core/placement.cpp
+++ b/form/core/placement.cpp
@@ -2,6 +2,8 @@
 
 #include "placement.hpp"
 
+#include <string>
+#include <string_view>
 #include <utility>
 
 using namespace form::detail::experimental;
@@ -20,3 +22,14 @@ std::string const& placement::file_name() const { return file_name_; }
 std::string const& placement::container_name() const { return container_name_; }
 /// Access technology type
 form::technology::id placement::technology() const { return technology_; }
+
+std::string form::detail::experimental::build_full_label(std::string_view creator,
+                                                         std::string_view label)
+{
+  std::string result;
+  result.reserve(creator.size() + 1 + label.size());
+  result += creator;
+  result += '/';
+  result += label;
+  return result;
+}
--- a/form/core/placement.cpp
+++ b/form/core/placement.cpp
@@ -2,8 +2,6 @@
 
 #include "placement.hpp"
 
-#include <string>
-#include <string_view>
 #include <utility>
 
 using namespace form::detail::experimental;
@@ -22,14 +20,3 @@ std::string const& placement::file_name() const { return file_name_; }
 std::string const& placement::container_name() const { return container_name_; }
 /// Access technology type
 form::technology::id placement::technology() const { return technology_; }
-
-std::string form::detail::experimental::build_full_label(std::string_view creator,
-                                                         std::string_view label)
-{
-  std::string result;
-  result.reserve(creator.size() + 1 + label.size());
-  result += creator;
-  result += '/';
-  result += label;
-  return result;
-}
--- a/form/core/placement.hpp
+++ b/form/core/placement.hpp
@@ -6,6 +6,7 @@
 #include "core/technology.hpp"
 
 #include <string>
+#include <string_view>
 
 /* @class placement
  * @brief This class holds all the necessary information to guide the writing of an object in a physical file.
@@ -35,6 +36,11 @@ namespace form::detail::experimental {
     /// Container name
     std::string container_name_;
   };
+
+  /// The container-name convention shared by writer and reader: creator and label joined as
+  /// "creator/label". FORM builds a placement's container name with it; the reader resolves the
+  /// same name back. Lives here, next to placement, so neither side owns the other's vocabulary.
+  std::string build_full_label(std::string_view creator, std::string_view label);
 } // namespace form::detail::experimental
 
 #endif // FORM_CORE_PLACEMENT_HPP
--- a/form/core/placement.hpp
+++ b/form/core/placement.hpp
@@ -6,7 +6,6 @@
 #include "core/technology.hpp"
 
 #include <string>
-#include <string_view>
 
 /* @class placement
  * @brief This class holds all the necessary information to guide the writing of an object in a physical file.
@@ -37,10 +36,6 @@ namespace form::detail::experimental {
     std::string container_name_;
   };
 
-  /// The container-name convention shared by writer and reader: creator and label joined as
-  /// "creator/label". FORM builds a placement's container name with it; the reader resolves the
-  /// same name back. Lives here, next to placement, so neither side owns the other's vocabulary.
-  std::string build_full_label(std::string_view creator, std::string_view label);
 } // namespace form::detail::experimental
 
 #endif // FORM_CORE_PLACEMENT_HPP
--- a/form/form/form_writer.cpp
+++ b/form/form/form_writer.cpp
@@ -29,9 +29,10 @@ namespace form::experimental {
 
   void form_writer_interface::parse_config(config::item_config const& config_item)
   {
-    // Parse the product configuration exactly once: map each product to its configured destination.
+    // Parse the product configuration exactly once: collect every configured destination for each
+    // product (a product may be written to more than one place).
     for (auto const& item : config_item.get_items()) {
-      config_by_product_.emplace(item.product_name, item);
+      config_by_product_[item.product_name].push_back(item);
     }
   }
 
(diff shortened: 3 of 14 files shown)


form/core/placement.cpp
```diff
@@ -2,6 +2,8 @@
 
 #include "placement.hpp"
 
+#include <string>
```
gemmeren: These should not be needed here, since they come from the header

form/form/form_writer.cpp
```diff
@@ -3,74 +3,105 @@
 #include "form_writer.hpp"
 
 #include <iostream>
-#include <stdexcept>
 #include <typeinfo>
+#include <utility>
```
gemmeren: Similar, already included in hpp.

form/persistence/persistence_reader.cpp
```diff
@@ -3,6 +3,8 @@
 #include "persistence_reader.hpp"
 #include "persistence_utils.hpp"
 
+#include "core/placement.hpp" // build_full_label
```
gemmeren: Maybe build_full_label should not be in placement.hpp?

gemmeren (dismissed):
Thanks Wanwei,
a few minor suggestions.

form/core/placement.hpp
```diff
@@ -35,6 +36,11 @@ namespace form::detail::experimental {
     /// Container name
     std::string container_name_;
   };
+
+  /// The container-name convention shared by writer and reader: creator and label joined as
+  /// "creator/label". FORM builds a placement's container name with it; the reader resolves the
+  /// same name back. Lives here, next to placement, so neither side owns the other's vocabulary.
+  std::string build_full_label(std::string_view creator, std::string_view label);
```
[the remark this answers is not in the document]
wwuoneway: `phlex::experimental` is the host framework's namespace, and FORM depends on phlex rather than the reverse, so moving FORM's naming helper there inverts the layering. This is consistent with the types it guilds names for `placement` and `token` — live in `form::detail::experimental`, so I'd keep it in `form::detail::experimental`.

form/persistence/persistence_writer.cpp
```diff
@@ -1,87 +1,93 @@
...
 
 using namespace form::detail::experimental;
 
+namespace {
+  // The navigation ("index") container lives alongside its product.
+  placement index_placement_for(placement const& product_place)
+  {
+    std::string const& name = product_place.container_name();
+    auto const slash = name.find('/');
+    std::string const creator = slash == std::string::npos ? name : name.substr(0, slash);
```
aolivier23, line 21: Please double-check that the compiler doesn't emit a warning about missing parentheses here.  I think it usually wants parentheses around the condition in a ternary operator.  I can read this personally and will approve as long as the compiler doesn't block us.

form/persistence/persistence_writer.hpp
```diff
@@ -6,43 +6,40 @@
 #include "ipersistence_writer.hpp"
 
 #include "core/placement.hpp"
-#include "storage/istorage.hpp"
+#include "storage/istorage.hpp" // brings in form/config.hpp (tech_setting_config)
```
aolivier23: Do we want to include `form/config.hpp` explicitly here too?  Some projects have an include-what-you-use rule.  I have to go look up whether that's in our coding guidelines.  But it helps prevent regressions if we update where `tech_setting_config` is visible one day.

aolivier23: Looks like it might eventually hit us through the c++ core guidelines SF.10

aolivier23 (dismissed):

form/form/form_writer.cpp
```diff
@@ -3,74 +3,105 @@
...
-    pers_writer_ = form::detail::experimental::create_persistence_writer();
-    pers_writer_->configure(config_item);
+  form_writer_interface::form_writer_interface(
+    config::item_config const& config_item,
+    config::tech_setting_config const& tech_config,
+    std::unique_ptr<form::detail::experimental::i_persistence_writer> pers_writer) :
+    pers_writer_(std::move(pers_writer))
+  {
+    parse_config(config_item);
     pers_writer_->configure_tech_settings(tech_config);
```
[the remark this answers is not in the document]
wwuoneway: Added a null check in that constructor that throws before configuration, so a null injection fails loudly instead of dereferencing null. thanks.

form/form/form_writer.cpp
```diff
@@ -3,74 +3,105 @@
...
+    using form::detail::experimental::placement;
+
+    auto const [plan_it, is_new_creator] = plans_.try_emplace(creator);
+    write_plan& plan = plan_it->second;
+
+    // ---- 1. PLAN ----
+    // Resolve every configured product to all of its placements and create those containers in one
+    // shot. FORM only names product containers; persistence adds the navigation ("index") container
+    // for each place itself, so FORM stays opaque to whether a place is indexed.
+    if (is_new_creator) {
```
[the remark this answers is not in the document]
wwuoneway: Good catch — the `is_new_creator` gate was too coarse. Fixed: resolution is now per product, not per creator. `write()` walks every product in the record and `try_emplace`s it into the plan, so each product is resolved the first time it appears (on whatever record that is) and its placements + commit key are built then. A configured product missing from the creator's first write is no longer skipped forever.

form/persistence/persistence_writer.cpp
```diff
@@ -1,87 +1,93 @@
...
-  for (auto const& [label, type] : products) {
-    containers.insert(std::make_pair(get_placement(creator, label), type));
+  std::map<std::unique_ptr<placement>, std::type_info const*> storage_containers;
+  for (auto const& [plcmnt, type] : containers) {
+    storage_containers.insert(std::make_pair(std::make_unique<placement>(plcmnt), type));
+
+    // Persistence owns navigation: every product container gets an index container alongside it.
+    placement index_place = index_placement_for(plcmnt);
+    auto const [it, inserted] = index_by_product_.try_emplace(
+      std::make_pair(plcmnt.file_name(), plcmnt.container_name()), index_place);
```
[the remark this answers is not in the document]
wwuoneway: Fixed — `index_by_product_` is now keyed by (file, container, technology), so two technologies into one file no longer collide on the same index.

form/core/placement.cpp
wwuoneway: Done - thanks.

form/form/form_writer.cpp
wwuoneway: Done - thanks.

form/persistence/persistence_reader.cpp
wwuoneway: Agreed. Confirmed it's used only to build container names (feeds `placement.container_name` on write and `token.container_name` on read, nowhere else), so it's fully self-contained. Splitting it into `core/container_naming.hpp` so the reader includes just that instead of all of `placement.hpp`. Thanks.

form/persistence/persistence_writer.cpp
wwuoneway: No warning with `g++` and `clang++`. `-Wparentheses` fires on nested ternaries and assignment-as-condition, not on a plain comparison as the condition, so nothing to change here.

form/persistence/persistence_writer.hpp
wwuoneway: Good point — the header uses `tech_setting_config` directly, so it now includes form/config.hpp itself rather than leaning on storage/istorage.hpp to pull it in.

gemmeren (approved):

PR #871, opened 2026-09-04
clang-tidy: resolve some warnings that slipped in

knoepfel:
Resolve `clang-analyzer-security` clang-tidy warnings.

This change silences a false positive that `clang-tidy` reports for the GCC build.  Ideally, we wouldn't need to make this change, but using the standard comment-based `clang-tidy` suppression mechanism is not successful because the false positive appears in an external Boost header file.

- **Code quality**
  - Resolve `clang-analyzer-security` warnings.
  - Replace iterator-based membership checks with C++20 `contains` calls.
  - Replace Boost’s `boost::is_any_of(":")` predicate with a lambda to avoid a false positive from an external Boost header.
  - Remove the related suppression comment.

- **Tests**
  - Pass `job_cursor` by value in the hierarchy traversal test.
  - Exercise the `data_cell_cursor` copy constructor before yielding the `"run"` child.

### DIFF
--- a/form/form/form_writer.cpp
+++ b/form/form/form_writer.cpp
@@ -85,7 +85,7 @@ namespace form::experimental {
       // appears at a place already written to, its container can no longer be added there -- fail
       // clearly here rather than crash deep in the backend.
       for (auto const& item : cfg_it->second) {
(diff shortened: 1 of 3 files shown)


wwuoneway (approved):
LGTM — thanks.

Issue #867, opened 2026-09-02
Accessor Functions for `root_rntuple_write_container` Public Members


aolivier23:
Move the public fields at https://github.com/Framework-R-D/phlex/blob/main/form/root_storage/root_rntuple_write_container.hpp#L61 into private fields with accessor functions.  The accessor functions will just provide a reference to `unique_ptr<>` for now because `root_rfield_write_container` needs to initialize these members.  These accessor functions have the bonus that we can update them with some extra initialization work in the far future.

aolivier23:
Solved by merging #880

PR #880, opened 2026-09-09
FORM RNTuple Accessor Functions

aolivier23:
RNTuple components that are used by multiple write containers are now `private`.  FORM containers manipulate them through accessor functions.  These accessor functions ensure that the RNTupleModel is converted to an RNTupleWriter the first time it is used.  This both solves issue #867 and sets FORM up for success with concurrent fills in the future.

- **Code**
  - Move shared RNTuple writer, model, and entry state into `root_rntuple_write_container_imp`.
  - Add `get_writer()`, `get_model()`, and `get_entry()` accessors.
  - Create the `RNTupleWriter` lazily through `RNTupleWriter::Append`.
  - Update FORM field containers to use the shared accessors.
  - Validate the configured `root_tfile_imp` and report invalid or missing files.
  - Commit the writer during destruction when it exists.
  - Prevent field commits before writer setup completes.

- **Concurrency**
  - Keep shared RNTuple components private to the write container.
  - Provide controlled access for multiple FORM field containers.
  - Support the planned concurrent-fill design.

- **Tests**
  - No test changes are reported.


aolivier23:
@phlexbot format

aolivier23:
@phlexbot format

aolivier23:
@phlexbot format

gemmeren (approved):
LGTM, Thanks Andrew.

wwuoneway (approved):
LGTM, thanks.

form/root_storage/root_rntuple_write_container.cpp
```diff
@@ -38,5 +51,31 @@ namespace form::detail::experimental {
...
+      writer_ = ROOT::RNTupleWriter::Append(std::move(model_), name(), *tfile_);
+    }
+
+    return *writer_;
+  }
+
+  std::unique_ptr<ROOT::RNTupleModel> const& root_rntuple_write_container_imp::get_model() const
+  {
+    return model_;
+  }
```
pcanal, line 70: Returning the implementation details (`std::unique_ptr`) is unusual.   Is that intentional or would returning a `ROOT::RNTupleModel*` be equivalent?

See Core Guide lines: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f7-for-general-use-take-t-or-t-arguments-rather-than-smart-pointers
and https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#r30-take-smart-pointers-as-parameters-only-to-explicitly-express-lifetime-semantics

Looking at the (current?) usage of the function, it looks like a simple function returning a bool (and having a more explicit name) would serve the same purpose.   Is there any use case for looking at the model before it is attached to the writer?

PR #893, opened 2026-09-15
Form write navigation layout

wwuoneway:
## What

Persist the write path's placement information as a navigation layout, so a later PR can locate any product directly instead of inferring its row from position in `creator/index`.

`register_write` already returns a real `token{file, container, technology, row}`; this PR accumulates that information and writes it at close:

- **`nav_<tech>_cells_<hierarchy>`** — one wide table per hierarchy per technology. One `uint64` column per layer, named as the cell index names it, plus one `<creator>_row` column per contributing creator. One row per data cell; a creator that never wrote that cell gets `invalid_row_id`.
- **`nav_<tech>_products`** — dictionary mapping each product to the navigation column that locates it: `product_name`, `creator`, `container_name`, `hierarchy_key`, `navigation_container`, `navigation_column`.

Navigation tables are ordinary FORM containers, one container per column, so no backend gained a new concept.

## Scope — write path only

Product bytes on disk are unchanged, `creator/index` stays as the reader's mechanism, and nothing under `form/storage/` or `persistence_reader.cpp` is touched. Existing read tests are unaffected.

- **Code**
  - Add structured `cell_index` types with layer names, values, consistency checks, and name sanitization helpers.
  - Change write and persistence APIs to accept `cell_index`.
  - Persist navigation layouts during `finalize()`.
  - Create technology-specific navigation cell tables and product dictionaries.
  - Validate hierarchy names, column collisions, duplicate cells, reserved prefixes, and invalid row mappings.
  - Make finalization explicit, idempotent, and safe during destruction. Reject writes after finalization.
  - Convert Phlex `data_cell_index` values into structured FORM cell indices.

- **Tests**
  - Add technology-independent navigation persistence tests.
  - Verify navigation rows map to the same data rows as `creator/index`.
  - Cover multiple technologies, hierarchies, products, invalid inputs, backend failures, finalization behavior, and post-finalize writes.
  - Update existing writer and storage tests to use `cell_index`.

- **Build**
  - Register navigation tests for each configured storage technology.


form/core/cell_index.hpp
```diff
@@ -0,0 +1,64 @@
...
+
+    auto operator<=>(cell_hierarchy const&) const = default;
+  };
+
+  /// Identifies a data cell by its canonical ID and layer coordinates.
+  struct cell_index {
+    /// Canonical cell ID as rendered by the framework.
+    std::string id;
+    /// Layer names, from outermost to innermost.
+    std::vector<std::string> layer_names;
```
gemmeren: should this be `cell_hierarchy`  ?

form/core/cell_index.hpp
```diff
@@ -0,0 +1,64 @@
...
+    /// Layer values corresponding to layer_names.
+    std::vector<std::uint64_t> layer_values;
+
+    bool is_job() const { return layer_names.empty(); }
+    /// Whether layer names and values have matching sizes.
+    bool consistent() const { return layer_names.size() == layer_values.size(); }
```
gemmeren: Do you even want to allow non-consistent `cell_index` or somehow force it to be consistent by design?

form/form/form_writer.cpp
```diff
@@ -31,6 +31,29 @@ namespace form::experimental {
     pers_writer_->configure_tech_settings(tech_config);
   }
 
+  form_writer_interface::~form_writer_interface()
+  {
+    // Safety net only; call finalize() explicitly to handle errors. Errors during destruction are
+    // reported but cannot be propagated.
+    try {
+      finalize();
```
gemmeren, line 39: Can Phlex call `finalize()` separately rather than relying on destructor? [I am not opposed to calling it from here, but having Phlex tell FORM to shut down may be preferable]

aolivier23 (dismissed):
LGTM modulo @gemmeren 's comments.  I want to see those addressed before merging please.

form/form_module.cpp
```diff
@@ -1,20 +1,47 @@
...
+      cell.layer_values.push_back(static_cast<std::uint64_t>(node->number()));
+    }
+    // Reverse to outermost-first layer order.
+    std::ranges::reverse(cell.layer_names);
+    std::ranges::reverse(cell.layer_values);
+
+    // Generate names for unnamed layers.
+    for (std::size_t i = 0; i < cell.layer_names.size(); ++i) {
+      if (cell.layer_names[i].empty()) {
+        cell.layer_names[i] = form::detail::experimental::unnamed_layer_name(i);
```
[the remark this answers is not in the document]
wwuoneway: Good catch — fixed,

form/form_module.cpp
```diff
@@ -45,6 +72,25 @@ namespace {
...
+      } catch (...) {
+        std::cerr << "form_output_module: finalize() failed with an unknown exception\n";
+      }
+    }
+
+    form_output_module(form_output_module const&) = delete;
+    form_output_module& operator=(form_output_module const&) = delete;
+    form_output_module(form_output_module&&) = delete;
+    form_output_module& operator=(form_output_module&&) = delete;
+
```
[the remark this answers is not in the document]
wwuoneway: Agreed. The hook is a Phlex change — `declared_output` is a bare `function_node`, so there’s no module end-of-job seam. I’ll file that upstream rather than change the module lifecycle in this FORM write-path PR; the destructor remains the safety net.

For the specific failure, the hook isn’t needed. `a-b` vs `a_b` is detectable when the second name is introduced. Both checks are now incremental via `claim_table_name` and `claim_column` in `record_navigation`, so invalid navigation throws from commit_place while the graph is running. finalize is now I/O only, and the tests assert the throws on the causing record.

form/persistence/persistence_writer.cpp
```diff
@@ -70,25 +108,271 @@ token persistence_writer::register_write(placement const& plcmnt,
...
+  try {
+    auto const it = index_by_product_.find(
+      std::make_tuple(plcmnt.file_name(), plcmnt.container_name(), plcmnt.technology()));
+    placement const index_place =
+      it != index_by_product_.end() ? it->second : index_placement_for(plcmnt);
+    store_writer_->fill_container(index_place, &cell.id, typeid(std::string));
+    store_writer_->commit_containers(plcmnt);
+
+    // Record navigation only after the product commit succeeds.
+    record_navigation(plcmnt, cell);
```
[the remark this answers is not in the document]
wwuoneway: Right on all counts, and fixed. 
`record_navigation` is now split. `stage_navigation` validates and computes the entire update — rows per creator, table-name claim, row novelty, new columns — without touching `navigation_tables_` or `claimed_table_names_`; it returns a `staged_record`. `commit_place` calls it first, commits storage, then calls `apply_navigation`, which does inserts only and cannot throw on validation.

test/form/form_navigation_test.cpp
```diff
@@ -0,0 +1,282 @@
...
+        }
+        checks.check(recorded == expected,
+                     "creator '" + creator + "' row " + std::to_string(rows[row]) +
+                       " holds the data cell navigation says it does");
+      }
+    }
+  }
+}
+
+int main(int const argc, char const** argv)
```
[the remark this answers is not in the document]
wwuoneway: The stated reason doesn't hold — Clang accepts `char const** argv`. Locally, `clang++ -std=c++20 -Wall -Wextra -pedantic -Wmain` with Apple Clang 21 compiles `int main(int const argc, char const** argv)` cleanly with no diagnostic.

I changed it to the standard `char* argv[]` form for consistency.

form/core/cell_index.hpp
wwuoneway: Yes — good catch. Fixed.

form/core/cell_index.hpp
wwuoneway: Good question — `to_cell_index` is the only production constructor and cannot produce a mismatch; other constructions are tests. Since `cell_index` is also FORM's public input to `form_writer_interface::write`, I think validating it at the boundary is preferable to enforcing the invariant in the type.

I also fixed a gap where `stage_navigation` returned before validation when there were no pending writes. It now validates every record.

form/form/form_writer.cpp
wwuoneway: Agreed, and I'd prefer that too — but Phlex can't today, which is why it's in the destructor.
`framework_graph::execute()` is `finalize()`; `run()`;, where that `finalize()` is graph construction and runs before any data. Nothing happens after `run()` returns. I'll open an issue to track this.

@coderabbitai please open a GitHub issue for the missing end-of-job hook, using the title and body below verbatim. Then link it here.

Title: No end-of-job hook for output modules: flush errors cannot reach execute()

Body:

`framework_graph::execute()` runs the graph and returns, but provides no end-of-job callback to registered modules. The `finalize()` in `execute()` is graph-construction finalization and happens before data flows.

As a result, output modules that need to flush buffered state (e.g. directories, indexes, or footers) must currently do so from their destructors. Since destructors cannot propagate exceptions, I/O failures during finalization can be caught and logged while execute() still returns successfully. This is currently reachable in FORM's output module.

Request: provide an end-of-job callback for modules that runs after run() completes and inside execute()'s error-propagation path, e.g.:

`form_output.at_end("finalize", &form_output_module::finalize)`;

The destructor should remain as a safety net for abnormal teardown, but normal finalization errors should propagate through `execute()`.

aolivier23 (approved):
LGTM

gemmeren (approved):

PR #897, opened 2026-09-16
feat: support 'stage' configuration parameter and API

### LINKED ISSUE #731: Configuration option to specify stage name
Users need to be able to specify the stage name in the configuration of the program.  This should be possible via:

1. A configuration parameter that can be in a configuration file, and
2. A command-line program option

The command-line option would take precedence over the configuration parameter in a configuration file.

knoepfel:
Fixes #731.

- **Configuration and CLI**
  - Added the `stage` configuration parameter.
  - Added the `--stage` command-line option.
  - Command-line values override configuration-file values.
  - Updated `phlex::detail::run` to pass stage and parallelism overrides.

- **Core API**
  - Propagated the stage identifier through `framework_graph`, `glue`, graph proxies, registration APIs, and node constructors.
  - Required explicit stage identifiers for `handle` and `product_store`.
  - Replaced optional product-store stages and the `"CURRENT"` fallback with required stage values.
  - Added `product_store::stage()` and updated product metadata propagation.

- **Tests and build configuration**
  - Updated unit tests for the new stage-aware APIs.
  - Added `--stage test` to relevant executable, benchmark, Python, plugin, and workflow tests.
  - Test execution results are not provided.


phlex/model/product_store.cpp
```diff
@@ -7,25 +7,21 @@
...
-                               phlex::detail::products new_products,
-                               std::optional<identifier> stage) :
-    products_{std::move(new_products)},
-    id_{std::move(id)},
-    source_{std::move(source)},
-    stage_{std::move(stage)}
+                               algorithm_name const& source,
+                               identifier const& stage,
+                               phlex::detail::products new_products) :
+    products_{std::move(new_products)}, id_{std::move(id)}, source_{&source}, stage_{&stage}
```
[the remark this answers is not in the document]
knoepfel: @coderabbit, for efficiency reasons, I do not want to store the `algorithm_name` and `identifier` by value in the `product_store`.  See the most recent commit, which purports to address this error by ensuring that the pointees persist onto all product stores have been destroyed.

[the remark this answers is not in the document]
knoepfel: @coderabbit, you wrote:

> The current API must then prevent rvalue metadata arguments, or otherwise require durable metadata at every construction site.

The former is not possible in C++ if the parameter types are of the form `T const&`.  I believe there is durable metadata at every construction site.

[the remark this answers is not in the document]
knoepfel: @coderabbit, I have adjusted the interface of `product_store` to take `gsl::not_null<T const*>` parameters, which makes it an error to directly provide a temporary object.  Please evaluate the most recent commit and see if this issue still needs further follow-up.

phlex/app/run.cpp
```diff
@@ -3,21 +3,33 @@
...
     return obj.at(key).as_object();
   } catch (std::exception const& e) {
-    throw std::runtime_error("Error retrieving parameter '" + key + "':\n" + e.what());
+    throw std::runtime_error(fmt::format("Error retrieving parameter '{}' :\n{}", key, e.what()));
   }
 }
 
 namespace phlex::detail {
-  void run(boost::json::object const& configurations, int const max_parallelism)
+  void run(boost::json::object const& configurations, overridable_configuration const& overrides)
```
marcpaterno: Consider a name change: overrides -> overridables; the original makes is seem that only the overriden parameters are contained.

phlex/core/registration_api.hpp
```diff
@@ -393,18 +405,18 @@ namespace phlex::detail {
...
-                                           resources_);
+        return std::make_unique<unfold_node<Object, Predicate, Unfold>>(
+          std::move(name_),
+          stage_,
+          concurrency_,
+          std::move(upstream_predicates),
+          graph_,
+          std::move(predicate_),
+          std::move(unfold_),
+          std::vector(inputs.begin(), inputs.end()),
```
marcpaterno: Consider going back to using std::from_range and moving `inputs`

phlex/app/run.cpp
knoepfel: See https://github.com/Framework-R-D/phlex/pull/897/commits/ef46e27e3eac1e95496a53b6bf32a2a937091f1c.

phlex/core/registration_api.hpp
knoepfel: See https://github.com/Framework-R-D/phlex/pull/897/commits/ef46e27e3eac1e95496a53b6bf32a2a937091f1c.

marcpaterno (approved):

PR #907, opened 2026-09-23
clang-tidy: fix configuration bug and apply changes

### LINKED ISSUE #904: Inline commits swallow rules in clang-tidy config
If I'm not mistaken, the incline configs in multi-line sections is split on commas, so that here:

https://github.com/Framework-R-D/phlex/blob/7b5ac2ef8437da0129531bea774d3b87390b7c3d/.clang-tidy#L21-L25

`# Use scripts/fix_header_guards.py instead misc-*, ` is parsed, so that the `misc-*` rules are never enabled and the following two lines become no-ops.

knoepfel:
Fixes #904.

---

**N.B.** This will be rebased after #911 has been merged.


plugins/python/src/pymodule.cpp
```diff
@@ -1,24 +1,20 @@
...
-
 #include "phlex/core/framework_graph.hpp"
-
+#include "phlex/module.hpp"
+#include "phlex/source.hpp"
 #include "wrap.hpp"
 
 #define PY_ARRAY_UNIQUE_SYMBOL phlex_ARRAY_API
-#include <numpy/arrayobject.h>
+#include <numpy/arrayobject.h> // IWYU pragma: keep
```
pcanal: Why does it confuses IWYU?

.clang-format
```diff
@@ -141,19 +141,23 @@ ForEachMacros:
   - BOOST_FOREACH
 IfMacros:
   - KJ_IF_MAYBE
-IncludeBlocks: Preserve
+IncludeBlocks: Regroup
```
pcanal: Would it be possible to separate the PR in the 'formatting only' change and the more functional changes?  In practice, I think this would mean one PR with the `.clang-format` change and its consequences followed by another PR with `.clang-tidy` change and its consequences.

knoepfel: Yes, that's possible.  I think, in this case, it will be easier to do the `clang-format` changes first, and then to do the `clang-tidy` ones.  I will restructure it.

plugins/python/src/pymodule.cpp
knoepfel: IWYU ends up identifying transitive headers in `numpy/arrayobject.h` which should be used instead.  However, the NumPy C API guidance is to use `arrayobject.h`.

.clang-format
knoepfel: See #911.  This PR will be rebased once that PR is merged.

PR #909, opened 2026-09-23
FORM RNTuple exception handling

aolivier23:
This PR unifies how FORM's RNTuple backend handles exceptions.  It is designed to wrap any place ROOT could throw an exception from RNTuple with a catch block.  These catch blocks add the FORM function name to the exception description and then throw a `std::runtime_error` with the next description.

- **RNTuple exception handling:** Added `handle_rexception` to include the calling function, operation context, and ROOT exception details in a `std::runtime_error`.
- **RNTuple reads and writes:** Routed ROOT exceptions from reader opens, reads, field setup, fills, commits, and appends through the shared handler.
- **Destructor errors:** Added function and file context to destruction-time commit errors while preserving the behavior of printing the message and allowing destruction to finish.
- **Build and formatting:** Added the handler source to the RNTuple build. Other changes reformat Python tests and code without changing their behavior.
- **Test status:** No test results were provided.


aolivier23:
@phlexbot format

aolivier23:
@phlexbot format

aolivier23:
@phlexbot format

wwuoneway (approved):
@aolivier23  thanks. The FORM RNTuple exception part LGTM.
I noticed some changes under `plugins/python/python/phlex` and `test/python/` that look like style or formatting changes. If that's the case, could you please check whether they may cause any merge conflicts.

PR #910, opened 2026-09-23
Remove unnecessary force_streamer_field configuration

wwuoneway:
**Summary**
- Remove the unnecessary force_streamer_field configuration in `test/form`. 
- Use the default RNTuple field mapping.
- Local test passes.

- **Code:** Removed the `force_streamer_field` overrides from the form writer and test configuration. Both now use the default RNTuple field mapping.
- **Tests:** The PR summary reports that local tests pass. No test results were independently verified.


aolivier23 (approved):
LGTM

